PR #6, opened 2015-07-13
Added '#include <string>' in misc.h for '__APPLE__' - fixes several errors when compiling on mac

lukecrowe:
This does not make it compile completely, but is a significant first step. May add more commits to this request if/when I determine the other needed changes.

Thanks!

### DIFF
--- a/opencog/util/misc.h
+++ b/opencog/util/misc.h
@@ -31,6 +31,10 @@
 #include <iterator>
 #include <functional>
 
+#ifdef __APPLE__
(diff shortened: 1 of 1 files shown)


lukecrowe:
I think this is more trouble than it's worth. Closing for now.

AmeBel:
Hi, you can setup using docker. See http://wiki.opencog.org/w/Building_OpenCog#Docker_.2864_bit_systems_only.29

PR #14, opened 2015-08-17
use include/opencog/util as install path

AmeBel:
The other option, that I know of, is to refactor all dependent project's preprocessor declarations. That maybe the 'right' thing to do, I don't know, or revert the project name change.


linas:
The need for this follows Handy's "obvious" pull request #13

PR #17, opened 2015-08-30
Fix crazy format crash

linas:
Some exception messages can contain the symbols %s %d %x %u etc and these can be interpreted as string formatting commands.  This in turn can lead to unexpected crashes. 

This happens in practice with guile exceptions, which can contains strings such as "%start-stack"  Note the %s in there...


ngeiswei:
That's an impressive bug finding. But I'm not sure the fix is at the right place. Shouldn't it be in Logger instead? So that, more generally, code like
```c++
logger().error(buf);
```
doesn't crash when buf contains format specifiers?

ngeiswei:
Never mind. I thought `buf` in exceptions.cc was a std::string, thus the bug was in Logger.cc rather than its use.

PR #33, opened 2016-01-07
Enables make install and make test under cygwin

mnuzzolilo:


tests/util/iostreamContainerUTest.cxxtest
```diff
@@ -31,28 +31,28 @@ using std::string;
 class printContainerUTest : public CxxTest::TestSuite
 {
```
mnuzzolilo: The changes in this file fix a bug related to the _N symbol being already used under cygwin.  Simply renamed N and _N it to NN and _NN respectively.

CMakeLists.txt
```diff
@@ -99,32 +120,12 @@ IF (CMAKE_COMPILER_IS_GNUCXX)
 	ENDIF (CMAKE_BUILD_TYPE STREQUAL "Coverage")
 ENDIF (CMAKE_COMPILER_IS_GNUCXX)
 
-# Do the windows build
-IF (WIN32)
```
mnuzzolilo: Win32 support is completely removed as of now.

CMakeLists.txt
```diff
@@ -157,7 +158,7 @@ IF(Boost_FOUND)
 	SET(BOOST_PARAMETER_MAX_ARITY 7)
 	ADD_DEFINITIONS(-DBOOST_PARAMETER_MAX_ARITY=${BOOST_PARAMETER_MAX_ARITY})
 ELSE(Boost_FOUND)
-	MESSAGE(FATAL_ERROR "Boost 1.46 or newer is needed to build OpenCog!")
+	MESSAGE(FATAL_ERROR "Boost 1.46 or newer (package libboost-all-dev) is needed to build OpenCog!")
```
mnuzzolilo: Clarifies error message to specify that libboost-all-dev is to be used. Took me awhile to figure that out myself

CMakeLists.txt
```diff
@@ -235,7 +236,8 @@ ENDIF (NOT DEFINED LIB_DIR_SUFFIX)
 # set default include paths
 INCLUDE_DIRECTORIES(${PROJECT_SOURCE_DIR} ${Boost_INCLUDE_DIRS})
 
-# rpath handling
+# RPATH handling (see https://cmake.org/Wiki/CMake_RPATH_handling)
+# Note: RPATH only supported under Linux!
 SET(CMAKE_SKIP_BUILD_RPATH	FALSE)
```
mnuzzolilo: Clarifies RPATH usage for anyone editing this script for cygwin or windows compilation

.gitignore
```diff
@@ -109,3 +109,8 @@ Makefile
 ### Linux ###
 *~
 *.autosave
+# Adding in-project build folders to git-ignore.
+# Any folder starting with the name build* is ignored
+# This is useful for cross-platform developers who might
+# want folders such as build-w or build-l for respective platforms
+build*/
```
mnuzzolilo: I use cogutils/build-w and cogutils/build-l for my build folders where cogutils is the project root.  It works well with my text/code editor, which only shows me files in the project root.  I don't foresee any non-build folders being created with the "build" string prefixed.

CMakeLists.txt
```diff
@@ -8,20 +8,41 @@
...
+# can override the path to the C and CPP compilers, respectively.
+# CMake will not update these settings internally until both the ./CMakeFiles
+# folder and ./CMakeCache.txt are deleted.
+
 include(lib/Summary.cmake)
 
-# Seems to work fine with cmake-2.6
-CMAKE_MINIMUM_REQUIRED(VERSION 2.6)
+CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12.2)
 IF (COMMAND CMAKE_POLICY)
```
mnuzzolilo: Updated CMake to 2.8 per the discussions on the group mailing list.

CMakeLists.txt
```diff
@@ -84,7 +105,7 @@ IF (CMAKE_COMPILER_IS_GNUCXX)
 	# 2) -fopenmp for multithreading support
 	#
 	# 3) -std=gnu++0x for C++0x and GNU extensions support
-	SET(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -Wno-variadic-macros -fopenmp -std=gnu++0x")
+	SET(CMAKE_CXX_FLAGS "${CMAKE_C_FLAGS} -Wno-variadic-macros -fopenmp -std=gnu++11")
```
mnuzzolilo: Update compilation to C++11

.gitignore
linas: can you add the missing newline?

CMakeLists.txt
linas: OK, but different distros have different naming conventions. That sounds like a debian/ubnuntu name.

CMakeLists.txt
linas: Hmm. I'm surprised. We shouldn't really be using RPATH, anyway.   RPATH is usually a hack intended for solving MacOS bugs.

CMakeLists.txt
mnuzzolilo: I'll clarify the message to state "if using Ubuntu" since most people seem to be using it.

CMakeLists.txt
mnuzzolilo: Yeah it seems to go against the cross-platform nature of CMake.  It definitely wasted a lot of my time trying to figure out why the RPATH commands weren't working under Cygwin

opencog/util/backtrace-symbols.h
```diff
@@ -10,12 +10,14 @@ extern "C" {
 char **oc_backtrace_symbols(void *const *buffer, int size);
 void oc_backtrace_symbols_fd(void *const *buffer, int size, int fd);
 #else
+#ifndef CYGWIN
 #include <execinfo.h>
```
linas: pleas don;'t place private nicludes into public header files. Nothing outside of backtrace-symbols.c needs execinfo.h therefore, this should not be placed in this header file. It just pollutes the namespace, and slows compilation.

linas: oops. never mind. deleted comment, I misread the change.

linas:
This looks fine to me (except for the missing newline).   I'll let Nil merge, though, as he has more interest in such stuff.

mnuzzolilo:
Okay made the corrections thanks for looking at it so quick!

CMakeLists.txt
linas: if you wish. it was an FYI comment.

CMakeLists.txt
linas: I'm pretty sure libcogutil does not need it.

But now that I think about it, the other libraries, e.g. `libatomsapce.so`, do need it, so that the path /usr/local/lib/opencog can be found, and otherwise, `ld.so` would not look there. (so its not for MacOS, but generally useful)

linas: actually, I'm confused. We should disable rpath and see what breaks

mnuzzolilo: I could do that in a subsequent pull request/merge.  As far as I can tell it's only used to locate the .so files from the executables. The unit tests should quickly determine the result

opencog/util/oc_omp.h
```diff
@@ -44,13 +44,16 @@
 ///@{
 
 //! compile with multithread support (!)
+#ifndef CYGWIN
 #define OC_OMP
+#endif
```
ngeiswei: I still don't understand why that is needed. I vaguely recall being able to run MOSES in parallel under Windows, the only difference would be that it was slower because thread creation under Windows is slower. But I might have imagined it, I'm not sure.

ngeiswei:
Thanks @mnuzzolilo !

PR #39, opened 2016-02-12
Build release package

wenwei-dev:


AmeBel:
I think the version should be update to one in https://github.com/opencog/cogutils/blob/master/opencog/util/cogutil.h. 

Maybe cpack macros should be dependent on cogutil.h macros or vice-versa. @linas what do you think?

linas:
Amen, don't over-engineer it.  There might be valid reasons to bump the atomspace version number, and leave the cogutils and opencog versions unchanged.

wenwei-dev:
Hmm, "For simplicity, I'll use the version defined in version.h for atomspace, and cogutil.h for cogutils and leave opencog as it is now, which is 0.1.4. "

CMakeLists.txt
```diff
@@ -303,9 +314,10 @@ SET(CPACK_SOURCE_STGZ "OFF")
 SET(CPACK_SOURCE_TBZ2 "ON")
 SET(CPACK_SOURCE_TGZ "OFF")
 SET(CPACK_SOURCE_TZ "OFF")
+SET(CPACK_PACKAGE_CONTACT "team@opencog.org")
```
linas: I don't think this mailing list exists. I think you want to use <opencog@googlegroups.com> as the contact point.

linas:
This looks reasonable (except for the bad email address) and can be merged. Wenwei, can you update the pull request?

PR #56, opened 2016-08-16
Changed async_caller to call const method

rTreutlein:
This is needed for https://github.com/opencog/atomspace/pull/861 and will be accompanied by a pull request that updates the AtomSpace to work with this change.


rTreutlein:
Should be merge together with, https://github.com/opencog/atomspace/pull/862.

PR #80, opened 2017-08-10
Build fix: changed std::abs() -> std::fabs() to be built with G++ 7.1.0

mhatta:
MOSES build with newer(?) G++ fails (see moses PR #49).  Changing abs() to fabs() fixes this.


ngeiswei:
Thanks.

PR #81, opened 2017-08-17
Debian packaging

AmeBel:
This uses to cpack to create debian packags. 
To build packages
```shell
mkdir build
cd build
cmake ..
make
make package
```
The package is found in `build/packages`. 
I have used `git describe --long` for versioning. 

Any suggestions on following this versioning approach?


AmeBel:
@mhatta what do you think?

mhatta:
I was not sure how I should name this Debian package -- The repo name is cogutils, but the library name is libcogutil.so (without "s").  Lintian suggests I should name the package "libcogutil", but this is clearly a devel package, so I called it "libcogutil-dev".

Also, AFAIK there is no clear rule for adding git versioning stuff.  I simply use the date and 7-digit hash so that I can easily compare it with GitHub "latest commit".  "~" and "+" are basically the same.

mjsduncan:
thanks for getting the packages together! maybe one thing to make clear is that debian buster or sid is needed for current package dependencies. another thing to reconcile is the use of /usr vs /usr/local.  for instance, bioscience in github/opencog/agi-bio uses /usr/local/[include, lib, share] for installing.
also there isn't an easy way to install npm for github/opencog/external-tools/glimpse.

mhatta:
@mjsduncan I think it's not difficult to rebuild my packages for stretch or Ubuntu.  Actually I can provide packages for stretch if needed (by using pbuilder).

As you said, /usr/local is not allowed for Debian packageing, but it doesn't mean you can't use /usr/local.  Under /usr/local is completely controlled by local admin.

It's a shame that npm is not well maintained in Debian, but I know it's tough...

AmeBel:
I had tested it in debian:jessie docker container and all is well for this repo and building atomspace using opencog/opencog-deps. 

I chose the output of `git describe --long` for versioning because it counts the number of commits since the last annotated tag. Thus, one can uniquely identify which revision is being used, even if there are multiple builds per day as a result of multiple merges.

linas:
Hey @mjsduncan The rules for `/usr` vs `/usr/local` is very simple: The distro (debian, redhat, etc) always installs to `/usr` and never to `/usr/local`  The local sysadmin (you, when you comile opencog from scratch) should almost always install to `/usr/local` and never to `/usr`.  

The idea is simple: you never clobber the distro, and the distro never clobbers you.  The actual rules  are a bit more detailed and subtle, and are given in the standards-body-approved LFSSTND the "linux file system standard'

mjsduncan:
i get the rationale, i'm just pointing it out so it can be documented to make using the packages as "it just works" as possible.  i sym-linked the bioscience binary to /usr/lib.  what do people suggest as the best/safest/most proper way to deal with this?

linas:
technically, hand-creating a symlink of some random binary to `/usr/bin` is an error. If you built the bioscience binary yourself, it should install into `/usr/local/bin` and work just fine there. If it doesn't, then I am guessing that `/usr/local/bin` is not in your search path?  It's supposed to be in there, and you should not need to do anything special to enable it.

PR #93, opened 2017-10-18
Fix is_disjoint and others

ngeiswei:
1. Fix `is_disjoint` for containers only implementing `operator<`.
2. Remove no longer useful `StringManipulator.h`. All these functions have STL or boost counterparts.
3. Rename `set_insertion` by `set_union_modify`, more meaning name.


AmeBel:
The removal of StringManipulator.h is breaking build in opencog repo

ngeiswei:
Sorry about that, can you paste the error. All functions in this file have std or boost counterparts as far as I remember.

ngeiswei:
Oh, I see it's just a missing include apparently.

PR #105, opened 2018-01-26
fix: while loop spinning on ; rather than subsequent loop

edguy3:
This pattern exists in opencog/util/async_method_caller.h code twice: 
```
while (...); 
{
   ... not a loop due to ; above
}
```
Probably not what was intended.  This patch fixes this problem. 
( Passes unit test on Ubuntu 16.04 ; no impact on atomspace tests )


linas:
wow. Good catch. The barrier and flush functions are rarely used,, but this would account for the higher-than expected cpu use when they are.

PR #106, opened 2018-01-26
type warning fix on OSX

edguy3:
```
.../cogutil/opencog/util/Logger.cc:422:60: warning: format '%ld' expects argument of type 'long int', but argument 5 has type '__darwin_suseconds_t {aka int}' [-Wformat]
                 "[%s:%03ld] ",timestamp, stv.tv_usec / 1000);
                                          ~~~~~~~~~~~~~~~~~~^
```
Added explicit cast to long to ensure proper processing.


linas:
thanks

PR #108, opened 2018-04-24
Add ThreadLocal utility class

vsbogd:
Add ThreadLocal utility class to allow threads using local instances of classes without contention.
https://github.com/vsbogd/atomspace/tree/parallel-pattern-matcher depends on it.

ThreadLocal usage example: https://github.com/vsbogd/atomspace/blob/cda8784ae7fdb754f2460157ae2cdf0049738212/opencog/query/DefaultPatternMatchCB.h#L156
Logger usage example: https://github.com/vsbogd/atomspace/blob/cda8784ae7fdb754f2460157ae2cdf0049738212/opencog/query/InitiateSearchCB.cc#L396


opencog/util/CMakeLists.txt
```diff
@@ -119,6 +120,7 @@ INSTALL(FILES
 	StringTokenizer.h
 	tbb.h
 	tree.h
+	ThreadLocal.h
```
ngeiswei: Total minor detail but there are some trailing spaces.

vsbogd: Fixed, thanks

ngeiswei:
@vsbogd You should update `Logger::set` so that `threadIdEnabled` gets properly assigned, I think.

opencog/util/Logger.cc
```diff
@@ -273,7 +274,8 @@ Logger::Logger(const std::string &fname, Logger::Level level, bool tsEnabled)
 }
 
 Logger::Logger(const Logger& log)
-    : error(*this), warn(*this), info(*this), debug(*this), fine(*this)
+    : error(*this), warn(*this), info(*this), debug(*this), fine(*this),
+      threadIdEnabled(false)
```
ngeiswei: Shouldn't `threadIdEnabled` just be assigned the value of `log.threadIdEnabled` instead of being initialized to false?

linas:
Can you explain this? Why is the standard thread-local storage, as supplied by standard c++, not sufficient to do everything you want?

linas:
Also, the logger is already thread-safe; so I am confused about why you feel you need a thread-local variant.

linas:
OK, now some comments re thread-local.

So, one standard idiom for storage that gets automatically cleaned up is to put that storage on the stack.  When the stack unwinds, the storage is cleaned up automatically.  I don't see why this is not sufficient in this case.

Separately, I kind of would have liked to talk to you about parallelizing the pattern matcher before you started writing code, instead of after. The "natural", simplest, easiest place to parallelize would be to make the initiate-search loops parallel, and then allow each search to run in a distinct thread.  This, it would seem to me, to be a natural place to set up the per-thread stack: its located "at the root" of the search, and so is guaranteed to be alive, until that thread finishes. 

Doing this would seem to obviate the need for some per-thread storage that shows up "from thin air", in the middle of some random code doing some random thing.  I'm rather nervous about the debuggability and correctness of what you are proposing to do. 

Anyway, this pull req is the wrong place to have this discussion.

vsbogd:
Extracted logging functionality to https://github.com/opencog/cogutil/pull/110, closing this PR

PR #109, opened 2018-05-08
Make RandGen instance thread local

vsbogd:
To use it correctly from multiple threads.

See discussion in https://github.com/opencog/atomspace/pull/1656


vsbogd:
Thanks, moved comments to header and fixed it.

ngeiswei:
:+1:

Issue #85, opened 2017-09-05
LoggerUTest fails


mhatta:
Currently, LoggerUTest got SegFault and fails on Debian sid.  8/31 snapshot (efa8e3b) can be built and pass the test in the same environment,  so recent commits broke something.

linas:
Did you do a `make; make install` first, before running the unit tests? Can you post a stack trace? There's no problem on Debian stable.

mhatta:
```
$ cd build/tests/util
$ ./LoggerUTest
unning cxxtest tests (6 tests)..[1]    6813 segmentation fault  ./LoggerUTest
```
```
$ gdb ./LoggerUTest
GNU gdb (Debian 7.12-6) 7.12.0.20161007-git
(snip)
(gdb) run
Starting program: /home/mhatta/work/Debian/opencog/opencog-cogutils/opencog-cogutils-2.0.3~git20170905.7b0b6f9/build/tests/util/LoggerUTest 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
Running cxxtest tests (6 tests)[New Thread 0x7f8808c3f700 (LWP 6846)]
[New Thread 0x7f880843e700 (LWP 6847)]
[Thread 0x7f880843e700 (LWP 6847) exited]
.[New Thread 0x7f8808c3f700 (LWP 6848)]
[Thread 0x7f8808c3f700 (LWP 6846) exited]
.[New Thread 0x7f880843e700 (LWP 6849)]
[Thread 0x7f880843e700 (LWP 6849) exited]
[New Thread 0x7f880843e700 (LWP 6850)]

Thread 1 "LoggerUTest" received signal SIGSEGV, Segmentation fault.
0x00007f880c626500 in _bfd_elf_find_function ()
   from /usr/lib/x86_64-linux-gnu/libbfd-2.29-system.so
(gdb) bt            
#0  0x00007f880c626500 in _bfd_elf_find_function ()                             
   from /usr/lib/x86_64-linux-gnu/libbfd-2.29-system.so                         
#1  0x00007f880c6025b8 in _bfd_elf_find_nearest_line ()                         
   from /usr/lib/x86_64-linux-gnu/libbfd-2.29-system.so                         
#2  0x00007f880d2b53d4 in find_address_in_section ()                            
[...]

PR #111, opened 2018-05-14
#85 - segmentation fault in debug mode on attempt to log stack trace

konstantin-t:
#85 is fixed.
Actually I noticed this fault when Opencog test failed in debug on my Ubuntu.


linas:
wow thanks.

PR #114, opened 2018-06-12
Assorted fixes

roman-khimov:
Various things found by static analysis and random jumps through the code. Probably wrong in some ways as I have absolutely zero knowledge of this codebase. Still, maybe there is something useful here for the project.


opencog/util/cluster.c
```diff
@@ -2922,10 +2922,7 @@ If a memory error occurs, pclcluster returns NULL.
   }
 
   /* Free temporarily allocated space */
-  free(data[0]);
-  free(mask[0]);
-  free(data);
-  free(mask);
+  freedatamask(nelements, data, mask);
```
ngeiswei: We unfortunately can't test since no more code is using that function, but that seems correct nevertheless.

opencog/util/exceptions.cc
```diff
@@ -100,11 +100,9 @@ StandardException::StandardException(const StandardException& ex)
 
 StandardException& StandardException::operator=(const StandardException& ex)
 {
-    message = NULL;
     if (ex.message)
     {
-        message = new char[strlen(ex.message) + 1];
-        strcpy(message, ex.message);
+        set_message(ex.message);
```
ngeiswei: Cool. It seems to me, you can apply the same simplification in the constructor, if you initialize message to `NULL` (or rather `nullptr`).

ngeiswei:
Thanks @roman-khimov . Do you think you could apply the simplification suggestion about the constructor, and then recreate a PR from a feature branch instead of the master?

opencog/util/exceptions.cc
roman-khimov: Yeah, I've thought about doing the same thing in the constructor but then decided to not touch it for some reason.

roman-khimov:
OK, let's try to do that.

PR #115, opened 2018-06-15
Various fixes

roman-khimov:
Some code cleanup, mostly for problems found with cppcheck.


opencog/util/exceptions.cc
```diff
@@ -128,9 +122,7 @@ const char * StandardException::get_message() const
 void StandardException::set_message(const char * msg)
 {
     // clear msg
-    if (message != NULL) {
-        delete [] message;
-    }
+    delete [] message;
```
linas: I guess its now OK to call delete on null pointers? Once upon a time, this would crash C++.

ngeiswei: According to SO it is in the standard https://stackoverflow.com/questions/4190703/is-it-safe-to-delete-a-null-pointer.

Issue #113, opened 2018-05-23
Organise OpenCog .cmake files


vsbogd:
Raising this issue here as it is across many opencog repos but most of the changes should be done in cogutil package.

1) .cmake find-module packages for 3rd party libs are copied and duplicated in few repositories. For instance [opencog/atomspace/lib](https://github.com/opencog/atomspace/tree/master/lib) and [opencog/opencog/lib](https://github.com/opencog/opencog/tree/master/lib) contains FindProtobuf, FindPython, FindGuile, etc modules. It leads to code duplication and necessity fixing  modules in each repository independently.

     All of such modules should be copied in single place. For instance we can keep them in opencog/cogutil repo, install them during cogutil installation and use installed versions from other repositories.

2. .cmake find-module packages for OpenCog modules (for instance FindAtomspace.cmake) should be located near the library code. 

    They should be kept in the repo which contains corresponding library and installed with the library. It makes sense to convert them into config-file packages (difference between find-module and config-file packages is described [here](https://cmake.org/cmake/help/latest/manual/cmake-packages.7.html#config-file-packages)) to make them compatible with 3rd party software which uses OpenCog as upstream.

vsbogd:
PRs https://github.com/opencog/cogutil/pull/120 and https://github.com/opencog/atomspace/pull/1832 fixed issue for opencog/cogutil and opencog/atomspace.
Repositories left:
- opencog/opencog
- opencog/moses
- opencog/benchmark

ngeiswei:
* opencog/as-moses

as well.

linas:
* opencog/agi-bio

as well

linas:
I'm sort-of changing my mind as to whether I like it or not. I see what's good about the proposal, and I like that -- it moves all similar files to one central location.   But there's also bad things  about this: the files move to somewhere far away from the users of the files. It makes sense if there are multiple users, and one provider.  But for many of these files, there is only one user (and there will only ever be one user) and so increasing the distance between user and provider does not seem like a good idea.

vsbogd:
Hi @linas, one argument I see is that when one is going to add new Find*.cmake tool he or she doesn't need search all repos and check if it is already here. Ideally these files should be a part of 3rd party library itself and it is not a part of code which uses it. So my idea was to move this files into utility library to allow using them from all other project if it is needed. ```cmake-data``` Ubuntu package does the same thing. It provides Find*.cmake scripts for third party libraries (SDL for instance).

[...]

PR #121, opened 2018-07-20
add Find*.cmake files from opencog(issue #113)

noskill:
As part of the work on issue #113 moving some cmake files from opencog


vsbogd:
looks good for me, but didn't tried to make clean build yet

vsbogd:
@linas , @ngeiswei, I think it is ready to be merged, if you don't have any objections I would merge it.

ngeiswei:
I didn't look at it but I trust your judgment, @vsbogd. Just wait for @linas' approval.

linas:
That is, I have no objections. I'm just feeling wishy-washy about it.

noskill:
@linas It would be better to provide config-file package for link-grammar too

vsbogd:
I have made some analysis and found that 4 of these libraries are actually not used by OpenCog and at least one should be converted to *Config.cmake actually. Below is the list of moved libraries with comments. Anyway I would raise additional PRs for these changes.

Not used:
- FindGSL.cmake - not used by opencog/opencog
- FindIODBC.cmake - not used by opencog/opencog
- FindLibMemCached.cmake - not used by opencog/opencog
- FindSDL_gfx.cmake - required by opencog/opencog (message is "SDL_gfx missing: needed for Spatial tools (MapExplorer, MapTools, etc.)." but looks like it is not used actually, and I cannot find sources for MapExplorer and MapTools only .dox files

Can be converted to *Config.cmake
- FindMOSES.cmake - MOSESConfig.cmake should be generated instead
- FindLinkGrammar.cmake - most proper way is to convert it into LinkGrammarConfig.cmake, but LinkGrammar itself is not cmake project

Used in opencog/opencog only:
- FindGHC.cmake - more or less universal it can be used by opencog/atomspace as it also builds Haskell bindings but it is not used in fact
- FindOctomap.cmake - used in opencog/opencog only
- FindUUID.cmake - used in opencog/opencog only
- Findcpprest.cmake - used in opencog/opencog only

vsbogd:
Thanks @noskill !

Issue #118, opened 2018-07-04
Loggers pointing to the same file step on each other


ngeiswei:
To reproduce the error comment out the following line

https://github.com/opencog/cogutil/blob/master/tests/util/LoggerUTest.cxxtest#L263

and run `testLoggerInteraction()`. The message of the first logger
will be cut before the end by the message of the second logger.

Till this get fixed the work around is to flush the first logger
before calling the second logger (as the test indicates).

ngeiswei:
Note that this issue is concerned about the garbage produced from messages stepping on each other, not about the loose ordering of the messages (which is also annoying but another problem).

PR #127, opened 2018-08-27
Work in progress for issue #118. Take 1.

mselseth:
- Added few error checkings.
- Fixed test code to execute correct checks, via TS_ASSERT under testLoggerInteraction(), /tests/util/LoggerUTest.cxxtest file
- Completed all paths to handle mutex while accesing logfile
- Added constructor method for Logger::Logger(const Logger& log)
- Need to follow up on some interactions between cout and logger
- Also need to fix white spaces and tabs next.

### DIFF
--- a/opencog/util/Logger.cc
+++ b/opencog/util/Logger.cc
@@ -169,8 +169,8 @@ void Logger::start_write_loop()
 void Logger::stop_write_loop()
 {
     {
-        std::unique_lock<std::mutex> lock(the_mutex);
         msg_queue.cancel();
+        std::unique_lock<std::mutex> lock(the_mutex);
         // rejoin thread
         writer_thread.join();
         writingLoopActive = false;
@@ -233,8 +233,9 @@ void Logger::flush()
 
 void Logger::write_msg(const std::string &msg)
 {
-    int   rv;
-    bool  continue_write = false;
+    int                 rv;
+    bool                error_detected = false;
+    std::ostringstream  error_description;
 
     // Note: stdout, stderr writing must be used with mutex unlocked
 
@@ -247,10 +248,14 @@ void Logger::write_msg(const std::string &msg)
         {
             logfile = fopen(fileName.c_str(), "a");
         }
-        if (logfile != NULL) continue_write = true;
     }
-    if (false == continue_write)
+    if (logfile == NULL)
     {
+        {
+            std::unique_lock<std::mutex> lock(sync_threads_mutex);
+            if (logger_message_count) logger_message_count--;
+        }
+
         fprintf(stderr, "[ERROR] Unable to open log file \"%s\"\n",
                 fileName.c_str());
         disable();
@@ -259,35 +264,46 @@ void Logger::write_msg(const std::string &msg)
 
     enable();
 
-    // Write to file.
+    /* Write to file is now protected by the mutex for entire fprintf and
+     * fflush operations, excluding the stderr operations
+     */
     {
         std::unique_lock<std::mutex> lock(the_mutex);
         errno = 0;
         rv = fprintf(logfile, "%s", msg.c_str());
-    }
-    if (rv != (int)msg.length())
-    {
-        fprintf(stderr,
-                "fprintf to %s file returned error: %s, "
-                "Detected %d bytes written, expected %d\n",
-                fileName.c_str(),
-                strerror(errno),
-                rv,
-                (int)msg.length());
-    }
+        if (rv != (int)msg.length())
+        {
+            error_detected = true;
+            error_description <<
+                "[ERROR] fprintf to " << fileName.c_str() <<
+                " file returned error: " << strerror(errno) <<
+                ", Detected " << rv <<
+                " bytes written, expected " << (int)msg.length() << std::endl;
+        }
 
-    // Flush, because log messages are important, especially if we
-    // are about to crash. So we don't want to have these buffered up.
-    {
-        std::unique_lock<std::mutex> lock(the_mutex);
+        // Flush, because log messages are important, especially if we
+        // are about to crash. So we don't want to have these buffered up.
         errno =  0;
         rv = fflush(logfile);
+        if (rv != 0)
+        {
+            error_detected = true;
+            error_description <<
+                "[ERROR] fflush returned error: " << strerror(errno) <<
+                std::endl;
+        }
     }
-    if (rv != 0)
+
+    /*  Synchronize the producer/consumer threads */
     {
-        fprintf(stderr,
-                "fflush returned error: %s\n",
-                strerror(errno));
+        std::unique_lock<std::mutex> lock(sync_threads_mutex);
+        if (logger_message_count) logger_message_count--;
+    }
+
+    if (error_detected)
+    {
+        std::cerr << error_description.str();
+        std::cerr.flush();
     }
 
     // Write to stdout.
@@ -393,7 +409,6 @@ void Logger::set_filename(const std::string& s)
 
     {
         std::unique_lock<std::mutex> lock(the_mutex);
-        if (logfile != NULL) fclose(logfile);
         logfile = NULL;
     }
 
@@ -499,6 +514,11 @@ void Logger::log(Logger::Level level, const std::string &txt)
 #endif
     }
 
+    {
+        std::unique_lock<std::mutex> lock(sync_threads_mutex);
+        logger_message_count++;
+    }
+
     msg_queue.push(new std::string(oss.str()));
 
     // If the queue gets too full, block until it's flushed to file or
@@ -512,6 +532,11 @@ void Logger::log(Logger::Level level, const std::string &txt)
     if (level <= backTraceLevel) flush();
 
     if (syncEnabled) flush();
+
+    while (logger_message_count != 0)
+    {
+        std::this_thread::yield();
+    }
 }
 
 void Logger::backtrace()
--- a/opencog/util/Logger.cc
+++ b/opencog/util/Logger.cc
@@ -150,7 +150,7 @@ Logger::~Logger()
 
     std::unique_lock<std::mutex> lock(the_mutex);
     if (logfile != NULL) fclose(logfile);
-	 lock.unlock();   
+    lock.unlock();
 }
 
 void Logger::start_write_loop()
@@ -161,7 +161,6 @@ void Logger::start_write_loop()
         writingLoopActive = true;
         writer_thread = std::thread(&Logger::writing_loop, this);
     }
-    
     lock.unlock();
 }
 
@@ -172,7 +171,6 @@ void Logger::stop_write_loop()
     // rejoin thread
     writer_thread.join();
     writingLoopActive = false;
-    
     lock.unlock();
 }
 
@@ -200,8 +198,8 @@ void Logger::writing_loop()
 
 void Logger::flush()
 {
-	 int rv;
-	
+    int rv;
+
     // There is a timing window between when pending_write is set,
     // and the msg_queue being empty. We could fall through that
     // window. Yes, its stupid, but too low-importance to fix.
@@ -215,17 +213,17 @@ void Logger::flush()
         usleep(100);
     }
 
-	 std::unique_lock<std::mutex> lock(the_mutex);
+    std::unique_lock<std::mutex> lock(the_mutex);
     // Force a write to the disk. Don't need to update metadata, though.
     if (logfile) {
-    	errno = 0;
-    	rv = fdatasync(fileno(logfile));
-    	if (rv != 0)
-    	{
-    	  fprintf(stderr,
-    	  			 "fdatasync returned error: %s\n",
-    	  			 strerror(errno));
-    	}
+        errno = 0;
+        rv = fdatasync(fileno(logfile));
+        if (rv != 0)
+        {
+            fprintf(stderr,
+                    "fdatasync returned error: %s\n",
+                    strerror(errno));
+        }
     }
     lock.unlock();
 }
@@ -234,7 +232,7 @@ void Logger::write_msg(const std::string &msg)
 {
     std::unique_lock<std::mutex> lock(the_mutex);
     int                          rv;
-    
+
     // Delay opening the file until the first logging statement is issued;
     // this allows us to set the main logger's filename without creating
     // a useless log file with the default filename.
@@ -257,9 +255,13 @@ void Logger::write_msg(const std::string &msg)
     rv = fprintf(logfile, "%s", msg.c_str());
     if (rv != (int)msg.length())
     {
-    	  fprintf(stderr,
-    	  			 "fprintf returned error: %s\n",
-    	  			 strerror(errno));    	
+        fprintf(stderr,
+                "fprintf to %s file returned error: %s, "
+                "Detected %d bytes written, expected %d\n",
+                fileName.c_str(),
+                strerror(errno),
+                rv,
+                (int)msg.length());
     }
 
     // Flush, because log messages are important, especially if we
@@ -268,9 +270,9 @@ void Logger::write_msg(const std::string &msg)
     rv = fflush(logfile);
     if (rv != 0)
     {
-    	  fprintf(stderr,
-    	  			 "fflush returned error: %s\n",
-    	  			 strerror(errno));    	
+        fprintf(stderr,
+                "fflush returned error: %s\n",
+                strerror(errno));
     }
 
     // Stdout writing must be unlocked.
@@ -336,7 +338,7 @@ void Logger::set(const Logger& log)
     this->printToStdout = log.printToStdout;
     this->printLevel = log.printLevel;
     this->syncEnabled = log.syncEnabled;
-   
+
     this->pending_write = false;
     this->writingLoopActive = false;
 
@@ -490,7 +492,6 @@ void Logger::log(Logger::Level level, const std::string &txt)
     // If the queue gets too full, block until it's flushed to file or
     // stdout. This can sometimes happen, if some component is spewing
     // lots of debugging messages in a tight loop.
-    
     if (msg_queue.size() > max_queue_size_allowed) flush();
 
     // Errors are associated with immenent crashes. Make sure that the
@@ -515,7 +516,9 @@ void Logger::backtrace()
     // If the queue gets too full, block until it's flushed to file or
     // stdout. This can sometimes happen, if some component is spewing
     // lots of debugging messages in a tight loop.
-    if (msg_queue.size() > max_queue_size_allowed) flush();
+    if (msg_queue.size() > max_queue_size_allowed) {
+        flush();
+    }
 }
 
 void Logger::logva(Logger::Level level, const char *fmt, va_list args)
--- a/opencog/util/Logger.cc
+++ b/opencog/util/Logger.cc
@@ -156,13 +156,11 @@ Logger::~Logger()
 
 void Logger::start_write_loop()
 {
+    std::unique_lock<std::mutex> lock(the_mutex);
+    if (!writingLoopActive)
     {
-        std::unique_lock<std::mutex> lock(the_mutex);
-        if (!writingLoopActive)
-        {
-            writingLoopActive = true;
-            writer_thread = std::thread(&Logger::writing_loop, this);
-        }
+        writingLoopActive = true;
+        writer_thread = std::thread(&Logger::writing_loop, this);
     }
 }
 
--- a/opencog/util/Logger.cc
+++ b/opencog/util/Logger.cc
@@ -255,8 +255,7 @@ void Logger::write_msg(const std::string &msg)
     // Write to file.
     errno = 0;
     rv = fprintf(logfile, "%s", msg.c_str());
-    
-    if (rv != (int)(msg.length() + 1))
+    if (rv != (int)msg.length())
     {
     	  fprintf(stderr,
     	  			 "fprintf returned error: %s\n",
--- a/opencog/util/Logger.cc
+++ b/opencog/util/Logger.cc
@@ -40,6 +40,7 @@
 #include <strings.h>
 #include <time.h>
 #include <unistd.h>
+#include <errno.h>
 
 #ifdef WIN32_NOT_UNIX
 #include <winsock2.h>
@@ -147,7 +148,9 @@ Logger::~Logger()
     flush();
     stop_write_loop();
 
+    std::unique_lock<std::mutex> lock(the_mutex);
     if (logfile != NULL) fclose(logfile);
+	 lock.unlock();   
 }
 
 void Logger::start_write_loop()
@@ -158,6 +161,8 @@ void Logger::start_write_loop()
         writingLoopActive = true;
         writer_thread = std::thread(&Logger::writing_loop, this);
     }
+    
+    lock.unlock();
 }
 
 void Logger::stop_write_loop()
@@ -167,6 +172,8 @@ void Logger::stop_write_loop()
     // rejoin thread
     writer_thread.join();
     writingLoopActive = false;
+    
+    lock.unlock();
 }
 
 void Logger::writing_loop()
@@ -193,6 +200,8 @@ void Logger::writing_loop()
 
 void Logger::flush()
 {
+	 int rv;
+	
     // There is a timing window between when pending_write is set,
     // and the msg_queue being empty. We could fall through that
     // window. Yes, its stupid, but too low-importance to fix.
@@ -206,13 +215,26 @@ void Logger::flush()
         usleep(100);
     }
 
+	 std::unique_lock<std::mutex> lock(the_mutex);
     // Force a write to the disk. Don't need to update metadata, though.
-    if (logfile) fdatasync(fileno(logfile));
+    if (logfile) {
+    	errno = 0;
+    	rv = fdatasync(fileno(logfile));
+    	if (rv != 0)
+    	{
+    	  fprintf(stderr,
+    	  			 "fdatasync returned error: %s\n",
+    	  			 strerror(errno));
+    	}
+    }
+    lock.unlock();
 }
 
 void Logger::write_msg(const std::string &msg)
 {
     std::unique_lock<std::mutex> lock(the_mutex);
+    int                          rv;
+    
     // Delay opening the file until the first logging statement is issued;
     // this allows us to set the main logger's filename without creating
     // a useless log file with the default filename.
@@ -231,11 +253,26 @@ void Logger::write_msg(const std::string &msg)
     }
 
     // Write to file.
-    fprintf(logfile, "%s", msg.c_str());
+    errno = 0;
+    rv = fprintf(logfile, "%s", msg.c_str());
+    
+    if (rv != (int)(msg.length() + 1))
+    {
+    	  fprintf(stderr,
+    	  			 "fprintf returned error: %s\n",
+    	  			 strerror(errno));    	
+    }
 
     // Flush, because log messages are important, especially if we
     // are about to crash. So we don't want to have these buffered up.
-    fflush(logfile);
+    errno =  0;
+    rv = fflush(logfile);
+    if (rv != 0)
+    {
+    	  fprintf(stderr,
+    	  			 "fflush returned error: %s\n",
+    	  			 strerror(errno));    	
+    }
 
     // Stdout writing must be unlocked.
     lock.unlock();
@@ -290,26 +327,27 @@ Logger& Logger::operator=(const Logger& log)
 void Logger::set(const Logger& log)
 {
     std::unique_lock<std::mutex> lock(the_mutex);
-    this->fileName.assign(log.fileName);
+
     this->component.assign(log.component);
     this->currentLevel = log.currentLevel;
-    this->printLevel = log.printLevel;
     this->backTraceLevel = log.backTraceLevel;
     this->timestampEnabled = log.timestampEnabled;
     this->threadIdEnabled = log.threadIdEnabled;
-    this->printToStdout = log.printToStdout;
-    this->syncEnabled = log.syncEnabled;
     this->logEnabled = log.logEnabled;
-
-    // Set NULL to force the logger to use a new FILE handle. It is
-    // safer that way because closing that file may not close file of
-    // the parent logger.
-    this->logfile = NULL;
-
+    this->printToStdout = log.printToStdout;
+    this->printLevel = log.printLevel;
+    this->syncEnabled = log.syncEnabled;
+   
     this->pending_write = false;
     this->writingLoopActive = false;
 
     lock.unlock();
+
+    // Set logfile to NULL to force the logger to use a new FILE handle. It is
+    // safer that way because closing that file may not close file of
+    // the parent logger.
+    // This action occurs under set_filename() below
+    set_filename(log.fileName);
 
     start_write_loop();
 }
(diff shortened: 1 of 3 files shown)


ngeiswei:
Cool! Will have a deeper review soon.

Meanwhile, @mselseth, could you fix the few indentations and trailing spaces minor formatting issues?

Also, minor remark, commit comments do not need to be conjugated to past tenses and do not need to end with a period. For instance
```
Corrected the error checking around fprintf call.
```
becomes
```
Correct the error checking around fprintf call
```
Thanks!

opencog/util/Logger.cc
```diff
@@ -147,7 +148,9 @@ Logger::~Logger()
     flush();
     stop_write_loop();
 
+    std::unique_lock<std::mutex> lock(the_mutex);
     if (logfile != NULL) fclose(logfile);
+	 lock.unlock();
```
ngeiswei: Weird indent

opencog/util/Logger.cc
```diff
@@ -206,13 +215,26 @@ void Logger::flush()
...
-    if (logfile) fdatasync(fileno(logfile));
+    if (logfile) {
+    	errno = 0;
+    	rv = fdatasync(fileno(logfile));
+    	if (rv != 0)
+    	{
+    	  fprintf(stderr,
+    	  			 "fdatasync returned error: %s\n",
+    	  			 strerror(errno));
+    	}
```
ngeiswei: Weird indent

opencog/util/Logger.h
```diff
@@ -160,7 +160,7 @@ class Logger
      * synchronous logging might slow down your program).
      */
     void set_sync_flag(bool);
-
+
```
ngeiswei: Trailing spaces

tests/util/LoggerUTest.cxxtest
```diff
@@ -246,15 +250,14 @@ public:
...
-        my_logger.set_timestamp_flag(false);
+        logger().set_component("LoggerUTest");
 
+        Logger my_logger(logger());
+        
         // Make sure both loggers point to the same file (not really
         // part of the test but better be safe than sorry)
         TS_ASSERT_EQUALS(logger().get_filename(), my_logger.get_filename());
-
+
```
ngeiswei: Trailing spaces

tests/util/LoggerUTest.cxxtest
```diff
@@ -267,9 +270,10 @@ public:
         my_logger.flush();
 
         std::string resline = getLastLineFromFile(my_logger.get_filename());
-        std::cout << "resline = " << resline << std::endl;
-
-        TS_ASSERT(resline == prefix + message or resline == prefix + gibberish);
+        std::cout << " resline length " << resline.length() << std::endl;
+        
+		  TS_ASSERT(resline.length() == (prefix.length() + message.length()));
```
ngeiswei: Weird indent

tests/util/LoggerUTest.cxxtest
```diff
@@ -52,12 +54,14 @@ public:
 
     static std::string getLastLineFromFile(const std::string& filename) {
         std::ifstream logfile(filename);
-        std::string line, resline;
+        std::string line = {}, resline = {};
         if (logfile.is_open())
         {
             while (logfile.good())
             {
+            	 line = {};
```
ngeiswei: Weird indent

opencog/util/Logger.cc
```diff
@@ -231,11 +253,25 @@ void Logger::write_msg(const std::string &msg)
...
+    }
 
     // Flush, because log messages are important, especially if we
     // are about to crash. So we don't want to have these buffered up.
-    fflush(logfile);
+    errno =  0;
+    rv = fflush(logfile);
+    if (rv != 0)
+    {
+    	  fprintf(stderr,
```
ngeiswei: Weird indent

opencog/util/Logger.cc
```diff
@@ -231,11 +253,25 @@ void Logger::write_msg(const std::string &msg)
     }
 
     // Write to file.
-    fprintf(logfile, "%s", msg.c_str());
+    errno = 0;
+    rv = fprintf(logfile, "%s", msg.c_str());
+    if (rv != (int)msg.length())
+    {
+    	  fprintf(stderr,
```
ngeiswei: Weird indent

mselseth:
Regarding the spaces, wrong indentations, I thing there are tools out there to fix the files.
Cannot remember  the name now, let me know if you have any suggestions.
I'll check the tools, and fix my editor as well. Used BlueFish for this pull request. Sorry about the indentations troubles. Yes, is really annoying.

mselseth:
For this repository, what are the code formatting rules?
tab indent expand to  2 spaces
convert tabs to spaces always
???

ngeiswei:
Here's the dev standard for OpenCog: https://wiki.opencog.org/w/Development_standards

Regarding automatic formating you may use astyle http://astyle.sourceforge.net/ . Here's an usage example for OpenCog https://wiki.opencog.org/w/Development_standards#Checking_with_astyle (beware that the options should be modified depending on the style of the file, space vs tab indentation, etc).

linas:
Do not convert tabs to spaces, ever. You can display tabs as 2 3 or 4 spaces - the choice is yours. I usually use 3 spaces.

opencog/util/Logger.cc
vsbogd: My opinion it is better to replace ```lock.unlock()``` by destructor call to make the scope of the lock more explicit. For instance replace three lines of code above by:

    {
        std::unique_lock<std::mutex> lock(the_mutex);
        if (logfile != NULL) fclose(logfile);
    }

opencog/util/Logger.cc
```diff
@@ -158,6 +161,8 @@ void Logger::start_write_loop()
         writingLoopActive = true;
         writer_thread = std::thread(&Logger::writing_loop, this);
     }
+    
+    lock.unlock();
```
vsbogd: ```lock.unlock()``` is not required here because it will be called by ```lock``` destructor.

opencog/util/Logger.cc
mselseth: Addressed.

opencog/util/Logger.cc
mselseth: Addressed

opencog/util/Logger.cc
mselseth: yes, the_mutex is part of class Logger
If I understand your comment, you don't think is needed to unlock as the Logger object destroys the mutex anyway.

I was concerned about this:
The behavior of a program is undefined if a mutex is destroyed while still owned by any threads, or a thread terminates while owning a mutex. 
from std::mutex manual.

Also, investigating if need for a try_lock logic.
Looking into this as well:
std::mutex is usually not accessed directly: std::unique_lock, std::lock_guard, or std::scoped_lock (since C++17) manage locking in a more exception-safe manner.

opencog/util/Logger.cc
mselseth: Addressed.

opencog/util/Logger.cc
linas: The mutex and the lock are not the same thing.  The mutex is in the class, and it is not destroyed until the class is destroyed. The mutex is shared by all threads accessing the calss.

The lock is a different C++ structure, and it is held only block scope:  it is created in the block, and destroyed when the block goes out of scope.  The lock is not shared between different threads: it just sits in the stack frame of the current subroutine call.

There is no need for try_lock, here. Try-locks are for solving various "thundering herd" and deadlock-related problems.  You don't have that here.

linas: Anyway, Vitaly knows this stuff. I am bowing out of this conversation, I've got too many other things to worry about.

vsbogd: Hi @mselseth, my comment is more about code style than about code correctness. As far as I can see your code is correct but explicit ```lock.unlock()``` is just not needed when ```unique_lock``` is used properly.

As @linas described above ```the_mutex``` is shared between threads and it is not destroyed while instance of Logger is live. ```std::unique_lock<std::mutex> lock``` is a helper which locks a mutex in constructor and unlocks it in destructor. The question is "why is this helper needed?" because you always can use ```mutex::lock``` and ```mutex::unlock``` methods explicitly. The answer is: by using ```unique_lock``` you can easily release the mutex correctly even when exception was thrown between ```mutex::lock``` and ```mutex::unlock``` calls (for instance in ```fclose``` call above).

To use this pattern one should define an instance of ```unique_lock``` at the first line of the code block which should be executed while mutex is locked. ```unique_lock``` locks the mutex in the constructor. And when control flow leaves the code block (it doesn't matter whether it is an end of the block reached or exception thrown) ```unique_lock``` destructor is called and mutex is released.

So taking this into account:
- if it is necessary to hold mutex only for the part of the method code (like in code snippet above) it is safer to create new code block and define ```unique_lock``` at the first line. Otherwise further code changes may lead to the situations when mutex is not released until the end of the method. It will not necessary happen but it is just a bit safer way of doing things.
- if mutex should be locked while whole method is executed it is enough to define ```unique_lock``` at the first line of the method. It is not necessary call ```unique_lock::unlock``` at the end of the code block (see my comment below https://github.com/opencog/cogutil/pull/127#discussion_r213958098)

mselseth:
I was wondering if anybody seen the errors about the config file, and what the remedy would be:
I get the following errors under tests/util/opencog.log (file logged when running make test), snippet below:
++++++++++++++++++++++++++++++++++++++
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=INVALID_PARAM
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=SERVER_PORT
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=SERVER_PORT
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=SERVER_PORT
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=LOG_FILE
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=LOG_LEVEL
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=BACK_TRACE_LOG_LEVEL
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=LOG_TO_STDOUT
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=SERVER_CYCLE_DURATION
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=EXTERNAL_TICK_MODE
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STARTING_STI_FUNDS
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STARTING_LTI_FUNDS
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STI_FUNDS_BUFFER
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=LTI_FUNDS_BUFFER
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=MIN_STI
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=PROMPT
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=MODULES
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STI_FUNDS_BUFFER
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=LTI_FUNDS_BUFFER
[2018-08-29 00:31:02:847] [INFO] Using config file found at: /home/marinela/work/code/git/cogutil/build/tests/util/./ConfigUTest.conf
ig

[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STI_FUNDS_BUFFER
[2018-08-29 00:31:02:847] [WARN] No configuration file was loaded! Param=STI_FUNDS_BUFFER
[2018-08-29 00:31:02:858] [INFO] Running LoggerUTest
+++++++++++++++++++++++++++++++++++++++++++++++++++++++

There is no ConfigUTest.config on my system.

Found these default conf files:
marinela@Inspiron-5566-Ubuntu:~/work/code/git/cogutil$ find . -name opencog.conf
./build/lib/opencog.conf
./lib/opencog.conf

mselseth:
Believe root caused this issue.

In the case of creating two  loggers, which is the case for the test that fails, there are separate concurrent_queue< std::string* > msg_queue; created under each logger.
So in this test, we end up with two instances of msg_queue, one per each logger object.

push to the msg_queue happens from the context of the same thread, test thread, which is serialized.
But each message ends up to separate queues.

pop of the queue occurs from the context of two newly created thread_writer threads.
Each thread_writer pops it's own queue.

Push always enqueues to tail, pop dequeues from head.
So with two queues, each queue has one message loaded.
The out of order occurs because the second thread_write thread ends up poping it's queue first, so that the writes to the file occur in reverse order.

Two possible solutions for this issue:
- share msg_queue, between multiple logger instances (1)
or
- introduce synchronization between the push code (producer) and the pop code (consumer), making the concurent_queue blocking (2)

I like option (1).

opencog/util/Logger.cc
```diff
@@ -147,26 +148,33 @@ Logger::~Logger()
...
+        std::unique_lock<std::mutex> lock(the_mutex);
+        if (logfile != NULL) fclose(logfile);
+    }
 }
 
 void Logger::start_write_loop()
 {
-    std::unique_lock<std::mutex> lock(the_mutex);
-    if (!writingLoopActive)
     {
```
vsbogd: It looks I have misleaded you a bit. If all method body should be protected by the mutex it is not necessary to create additional code block. So here it should be just as ut was:

void Logger::start_write_loop()
{
    std::unique_lock<std::mutex> lock(the_mutex);
    if (!writingLoopActive)
    {
        writingLoopActive = true;
        writer_thread = std::thread(&Logger::writing_loop, this);
    }
}

opencog/util/Logger.cc
```diff
@@ -147,26 +148,33 @@ Logger::~Logger()
...
-        writingLoopActive = true;
-        writer_thread = std::thread(&Logger::writing_loop, this);
+        std::unique_lock<std::mutex> lock(the_mutex);
```
vsbogd: vsbogd 2 minutes ago • Member

It looks I have misleaded you a bit. If all method body should be protected by the mutex it is not necessary to create additional code block. So here it should be just as ut was:

void Logger::start_write_loop()
{
    std::unique_lock<std::mutex> lock(the_mutex);
    if (!writingLoopActive)
    {
        writingLoopActive = true;
        writer_thread = std::thread(&Logger::writing_loop, this);
    }
}

mselseth:
commit #5 addresses the issue by synchronizing the producer thread and the consumer thread when multiple loggers are used.
Recommend diffing between original and this commit as I cleaned up some code.

Using one queue is more complicated solution, so I decided to go the synch route.  This kind of defeats the creation of separate thread for writing to the logger. After this change some critical path may be affected as the code will block to keep up with the logger writes, maybe can add this as a feature to the constructor.
So both ways are available to the user.

mselseth:
Added the block in case future code is added under this function, but not a big deal. commit #6 reverted it back.

linas:
@mselseth is right; if you want in-order messages, then you MUST use a single queue.

It is either impossible or insanely foolish to do it any other way (depending on your design choice).

The whole point of having queues in the first place was to speed up execution.  Without queues, each logger message is a `fprintf` to some file; this requires a system call (a call into the OS kernel) which will cause all deferred interrupts to run, it will possibly/probably cause a context switch to happen, as well as lots of other things ... fprintf/fwrite is SLOW.   With queues,  a message can be placed (very quickly) into an in-RAM queue, without any kernel overhead at all, and then return to the caller. 

If you always force writing to be synchronous, you get zero performance advantage (since you now pay for the kernel call) plus you get a performance disadvantage (pointless overhead of queing, plus a context switch is now always forced, to switch threads.  Please note that the Intel spectre/meltdown bugs makes context switching about 50% slower...

So its utterly, completely pointless to do queueing, if you are always going to run in synchronous mode. it is a 100% bad design.  Don't do it.  You have three choices: different loggers must use the same queue, or different loggers write to different files, or you accept that messages might be written out-of-order.

linas:
Note also: if you just want to run in synchronous mode, no new code is needed. There already is a logger flag that says "run in synch mode".

ngeiswei:
@linas, this PR is a WIP for issue #118. This issue is real and annoying, so any fix in that directly is welcome.

Just to be sure, out-order messages is not the problem, the problem is intra-message cut (i.e. screwed-up message).

I didn't have time to review @mselseth's work but I kinda suspect you're closing it prematurely.

@vsbogd, any thought?

(will eventually review it, sorry, didn't have the time/priority).

mselseth:
@linas thanks for jumping in, for being considerate.

Few clarifications.

1. The fix added under commit #5, **fixes** the problem, the messages are printed in-line and not interleaved.

The way it works it blocks under the user thread (the one that produces the messages), until the newly created threads handle the fflush to the logger, either one.
So it fixes the issue, but this blocking then defeats the original async design that creates new threads to consume the msg queue, to then log messages polled from queue.

This code accomplishes the blocking:
    while (logger_message_count != 0)
    {
        std::this_thread::yield();
    }

2. There are some other minor fixes under this branch, like:
- msg_queue already has mutex that protects queue access, internal to concurrent_queue class, so there is no need to protect queue access via mutex 'the_mutex' from Logger class.
the_mutex from Logger class should probably be only used to protect logfile access 
- added error checks when accessing logfile
- divorced logging to stderr from Logger mutex
- test code fixes

3. Yes, agree, the glorified solution is to have one single queue, one single write_thread started, but this fix is changing the design quite a bit.
It is my preferred solution, as explained when root cause identified.
Added the synch solution as a proof of concept, maybe jumped the gun when I published it.

4. Believe I've seen messages interleaved as well, not only out of order, when testing original code.

mselseth:
I meant thank you @ngeiswei in previous comment.

linas:
OK. Grmmmph. Do not ever call the correct solution the "glorified solution". Its not glorified; its the correct solution.  What you are proposing instead is an incorrect, broken solution to an ugly problem. If its good enough for Nil, I guess maybe just this once we can temporarily allow it. But still, I don't like it.

Committing broken code is almost always a really bad idea.  Why?

1) It reduces the pressure to create the correct solution.  If you don't have the time to do the correct solution now, when will you have the time? Tomorrow? Day after tomorrow? Never?

2) It raises the confusion for future maintainers. They see this complex code --  they figure that "it does something important", and they leave it alone.  Do have the time to figure out that it does something bad?  How much longer will it take them to figure out that the code is wrong, and that it should be removed? How much longer will it take them to find the right solution in the future?  

By committing something broken now, you are making things a LOT more complex for future maintainers.  It becomes very expensive to fix. But since you already see what the correct solution is, now, the best time to implement the correct solution is ... now.  Doing the right thing now avoids "technical debt".

https://en.wikipedia.org/wiki/Technical_debt

Don't create technical debt.  We can't afford it.

linas:
If you have other, unrelated minor fixes, pleas submit those in a distinct pull request.

mselseth:
Don't believe this code should to be merged yet.
At the time this pull request was closed it was a work in progress.
I didn't realize that anybody can close a wip pull request.
Was hoping for collaboration not arguments.

As I said twice already, the fix is to have unique queue, unique writer_thread.
Seems to me we agree on this design change.

I am ok with stopping my contribution at this point.
At least I identified the root cause, which is always half the battle.
My time is also important.

BTW, this is my first attempt to contribute to open source. I have to admit, it is not what I envisioned.

linas:
Ah. Look, normally, pull requests are for finished work, not for work-in-progress. When you ask someone to pull your code, there is an implicit understanding that you are done with it. ("pull requests" are not "requests for collaboration" or "requests for comment"; if they were, that is what they would be called)

Collaboration in open source is the same as collaboration in closed source (or, at least, should be, if you are working with a well-managed proprietary project). The goal is always to create high-quality constructions. 

Please do not think of this as "an argument". This is what a collaboration actually looks like. My goal here is to show you how to actually do things, how to get things done.  Do not think that the act of creation is effortless or easy. It takes work and time and effort and skill and practice.  When someone tries to show you how to do something complex, do not think of it as "arguing".  Its just complex. Its not easy.

No one said your time is unimportant. I am investing my time, in you, as a teacher to a student, hoping that one day you can achieve great things. If you want to be a bad student, and run away, well, that just hurts me, and makes me less hopeful about the next student. 

Do the right thing, here.  Create one or more distinct pull requests containing the various minor fixes and cleanups. Those should not be contentious, and can be easily merged.  The more difficult and complex design ... well, please do try to start it. Maybe it will be found to have other difficulties or problems.  Some of the difficulties and problems do not become visible, until you attempt a design.

FWIW, I throw away two or three designs for every one that I keep.

ngeiswei:
I wonder, what about making `Logger::write_msg` atomic across loggers sharing the same destination?

As a quick experiment I tried to make `the_mutex` static but got a compile error.

Any thought?

(@mselseth we appreciate your help and you're definitely welcome here. Lack of concern for good manners seems to be almost the norm in the FOSS community as the technical aspects take precedence over everything else, see Linus Torvalds and Richard Stallman, it says it all!).

mselseth:
@ngeiswei , not sure it will solve the problem; the pop still have the potential of happening out of order, unless the_mutex is shared across all threads including the user thread; but this in turn can potentially create deadlock scenarios, when you share a mutex across multiple threads.
Once the pop is out of order, downstream code that writes will occur out of order.

The reason there is no deadlock now is that there is mutex (the_mutex) for each thread, which kind of defeats the mutex purpose.

Not sure about the compile error. What do you get?
I had compile issues as well when I attempted to implement solution #1.

+++
Going forward with this, I can remove the synch code from this branch, cleanup the minor fixes so they can be merged.
Please re-open this pull request, don't think I can do it myself, and will publish the changes minus the synch code that blocks.
Should make review simpler.

Then, I'll create a new pull request/branch for addressing solution #1.

jacleland:
Hi Marinela,

I thought I would share my experience with you regarding this design
pattern. I apologize for jumping into the discussion uninvited. After
reading some of the messages, I'm of the impression that, essentially, each
child thread has a queue and a mutex, is this correct? One thread (user?)
is responsible for popping message from each child thread's queue and
writing them to the actual log file? If this is the case, I don't believe
you would need to be concerned about deadlocks as only the user and child
threads are being serialized while accessing each child's queue.

Admittedly, I'm NOT looking at the logger right now, but unless there's a
"head-on" condition where two or more threads are attempting to lock 2 or
more semaphores in _different order_, I don't expect you'll introduce the
potential for deadlocks given the design parameters described above. If I'm
misunderstanding the implementation and there is a single mutex serializing
access to either a single queue or all child queues, the
design/implementation might want to be reviewed. This would essentially
serialize all child threads and the logging thread which seems rather
inefficient.

I believe Linas mentioned this as well, but my understanding from the
original explanation was that the per-thread queuing of log messages was
being bypassed (to insure [time] proper ordering of log messages?) and each
child thread was essentially locking a single mutex for the purpose of
serializing access to the log file handle, and then calling the RT lib to
write it's own data to the log. As was mentioned before, this would add
considerable overhead not just at the single mutex shared by all threads,
but by forcing a context switch every time a child thread makes a RT call
into kernel space.

I'm new here as well and currently familiarizing myself with the software.
The implementation that I described initially is familiar to me however,
and has always performed well. The tradeoff, as Linas mentioned, is that
typically log messages are popped from each queue (perhaps in a circular
fashion and one at a time) and end up written to the log file accordingly
and without regard for when the child thread actually _logged_ the message
or any timestamp that may be included in the log message data. If you
happen to be stuck on this issue, then I hope the above might help you out.
Regarding the defect involving interleaved log messages, that would require
a closer look. :)

Good luck!

mselseth:
Yes, for test testLoggerInteraction, there is the user thread, and two threads writer_thread, one from each logger, one msg_queue for each logger, call it Q1, Q2.
The push happens from user thread, it is serialized, but it happens to each logger object, so it uses the msg_queue that corresponds to each logger message, Q1 for large message, Q2 for small message.
Once queues are populated, they are poped in the context of writer_thread. Each writer_thread runs asynch from everybody else.
The poping of the message to be written, occurs in the context of writer_thread, not user thread. This is important detail.

Hopefully, this explanation adds more clarity, but it is no different from my original assessment.

I would not worry about deadlocks yet, until we get to the change with one queue, or other static mutex changes.

linas:
Hi @jacleland 

The original logger, from way back when, had one single thread-safe queue that any writer could write to at any time, and the queue would automatically "do the right thing" and queue up those messages.  On the back-end, there's a thread that drains the queue and writes the messages to a file.   The queue itself is designed to be fully thread-safe, so that anyone anywhere (any thread) can push onto it at any time, and anyone anywhere can pop/deque from it at any time, and it all just automagically works.  And its fast.

The problem is that ... someone started using TWO loggers.  Which is the first design mistake.  The second design mistake was the desire to have both loggers write to the same file at the same time.  If you know anything at all about how file systems work, you would immediately know that the second design mistake is a fatal flaw, and that you will get garbage in the file. User error.  So that's what we've got.

So now we are trying to figure out how to get out from under two fundamental design errors. I'm frustrated because, instead of recognizing the design flaws as flaws, and actually fixing them, the proposed solutions are to double-down, and add random hacks to random places, experimenting with the hacking until some almost-solution is accidentally stumbled upon.  This is not the correct way to do engineering. 

Fix the original design errors. Then the bugs will go away.

mselseth:
Having one logger per process, across multiple user threads seems reasonable to me as a design decision.
The logger code could make the logger truly singleton, with no way to  create another logger. I think this can be enforced via constructors.
Also, from a design perspective, don't know why timestamp is optional.
A logger should probably always record the ts.
An out of order message can be easily spotted via the ts, in this new design should not happen.

jacleland:
Thank you for taking the time to provide this explanation, Linas. This
background information goes a long way in terms of helping me to understand
the logging subsystem. You've also given me something to think about
regarding the interleaved writes that were mentioned yesterday and how this
defect may have been introduced. Cheers!

linas:
1) Actually, you can have multiple loggers per process, as long as they do not all write to the same file. So it does not have to be a singleton.

2) The timestamp is optional because it takes up a lot of room in the log, making it hard to read.  When you actually have to actually use the log in real life, to debug something, the time-stamps are super-duper annoying, as one really does not care about them -- one wants to know the actual log messages, only. The time-stamps use up columns 1 through 35, filling those columns with pointless junk.

3) Detecting out-of-order timestamps is really really hard for human beings. It is eye-wateringly dull and difficult to stare at hundreds or thousands of timestamps, attempting to spot one or two out of order, while also trying to understand the actual messages.  Debugging from log messages is already hard. Trying to mentally reshuffle the log messages is even harder.

mselseth:
Since logger is utility for applications to use, I would think a process with multiple user threads would most likely want to use one logfile.
The trace in the logfile can be used to debug threads interactions so the order of messages, and their integrity is important, especially when the timestamps are optional.
Personally, I like to see how/when threads execute and interact, under one logfile, for timing and other non-repeatable issues.

Creating separate logfiles, yes, would mask the problem, but I believe having one logfile from different user threads is valid use case, because it adds context.

Test testThreadIdLogging just creates a simple, wonderful repeatable scenario to uncover a design issue.

ngeiswei:
I'm the one who introduced the multiple logger thing. I did that because I wanted to have the logger support multiple components (rule engine, etc) so that each component can easily tweak their logging parameters (level, stdout output, component name, etc). This is incredibly useful and I don't want to remove these functionalities. Supporting this with only one logger as presently designed is inconvenient, each message would need to change the logger parameters before logging, then restore the previous parameters, not good.

Then regarding logging on the same file, it allows to see the flow of information from components to components, I don't want to remove that either.

Perhaps the right way to do that is to have one logger instance supporting multiple settings/components. So the user would always invoke the same `logger()` singleton but would pass an optional component argument to use settings deviating from default. In a way it seems we're converging to that  (single queue, etc). So, maybe we need to step back and think about it...

One the other hand I do suspect an easy fix/hack could remove this bug and buy us time to before we need to bite the bullet and redesign a better multi-component support.

P.S: I'll come back soon with more on the PR itself.

opencog/util/Logger.h
```diff
@@ -315,6 +315,9 @@ class Logger
     bool syncEnabled;
     FILE *logfile;
 
+    std::mutex sync_threads_mutex;
+    unsigned int logger_message_count = 0;
```
ngeiswei: Please add comments about these new members.

opencog/util/Logger.cc
```diff
@@ -461,6 +530,11 @@ void Logger::log(Logger::Level level, const std::string &txt)
     if (level <= backTraceLevel) flush();
 
     if (syncEnabled) flush();
+
+    while (logger_message_count != 0)
+    {
+        std::this_thread::yield();
+    }
```
ngeiswei: OK, I finally get the idea. But I don't think this is acceptable as it essentially makes `Logger::log` a blocking call, which we don't want to efficiency reason. And, as Linas said, we already have the `sync` flag for that.

ngeiswei:
@mselseth after carefully reviewing your code I cannot accept it for the same reason that @linas puts forward, namely it turns `Logger::log` into a synchronous call.

However it does contain good stuff in it, like the extra checks and so on (which should be observable by disabling the flushing in `testLoggerInteraction()`, I suppose that is what drove you to write them). So if I would recommend that you create a new PR with these only (i.e. without `sync_threads_mutex`, etc), and strive to make it as clean as possible, that is having each commit capture to a self-contained idea (insofar as possible) and the PR constituted of a cohesive stream of commits (i.e. no mutually cancelling out changes, such as a commit introducing wrong indentation, followed by another correcting it, etc).

By using either git CLI (`git add -p` if I'm correct) or any git front-end your IDE might offer (I personally use magit, an awesome git front-end for Emacs), you can craft your commits as cleanly as possible by choosing which hunk to include, how large should the hunks be, etc.

Thanks for your interest and contribution so far.

linas:
Look, I think I explained it really quite clearly at least twice, I'm not sure why I have to explain it a third time. You can get everything @ngeiswei nil wants, if you use one queue.  Just use one queue, and everything else works automatically.

mselseth:
When I identified the root cause suggested two solutions, nobody said anything at the time, so I went ahead and published the sync code.
Already agreed to remove the sync code. This is not the discussion we need to debate.
The PR was closed before I could publish the subsequent fixes.

Still don't understand why not fix everything under this PR, if it can be re-opened, or have the minor fixes published under this PR (need to cleanup code of course) and use a separate PR for the single queue re-design.
Can anybody explain to me some PR etiquette?
This is what is holding me. Yes, I can open a new PR, what is the guarantee it won't be closed again. if somebody doesn't like something, instead of allowing me to follow up?

From my experience with git tool, you can squash a PR that holds multiple commits, when is merged, so the fix can be picked up on the master as a single commit from this branch.
This is why I forked a repo and created separate branch.

ngeiswei:
@mselseth, that PR contains a lot of stuff, most of which is not directly pertinent to solving this problem, which is why I suggest
1. Create a PR containing your cleaning/robustness improvements only
2. then create another PR containing the fix for the issue.

If you fix order+garbage with a single queue, great, if you fix only garbage (I mean messages stepping on each other), great too (that's what I care the most, personally).

Maybe @linas prematurely closed the issue, on the other hand, I do feel that it's better to start from a fresh PR rather than bloating that one any further, especially because that way you can split it in 2, cleaning/improvements vs fixing #118). I give you the guaranty that if your PR looks great it won't be closed ;-).

linas:
Look, I think I explained it really quite clearly at least twice, I'm not sure why I have to explain it a third time. You can get everything @nil wants, if you use one queue.  Just use one queue, and everything else works automatically.

linas:
I just created pull re #129 that I beleive solves the problems being encountered here.

It seemed easier to just create the fix than to try to explain it.   

I have not merged the code.  Please take a look at what it does. Its not a complicated fix -- it associates each distinct file-name with one queue.

Issue #126, opened 2018-08-22
Find* cmake files installation path variable


Dagiopia:
currently it's installed to DATADIR/cmake but it's better to use CMAKE_INSTALL_PREFIX/cmake because during compiling the intended data path and the installation path may be different. when building packages and crosscompiling this is a problem. I think it's the entire point of having CONFDIR and DATADIR they're usually similar to CMAKE_INSTALL_PREFIX but they get embedded in C source files and the path is needed to be right on the target system rather than the host. 
https://github.com/opencog/cogutil/blob/f1aced35095180f78e68058b1223876c393da10d/cmake/CMakeLists.txt#L41

noskill:
Currently DATADIR is used to install *.scm files, to install cmake files, it also used from c++ to load scheme and python files... So should installation path be updated for scm files too? Is it really broken during crosscompilation?

Dagiopia:
Actually it doesn't really break anything. It could be driven around but DATADIR path is as you said for the scheme files which makes what's done [here](https://github.com/opencog/atomspace/blob/master/opencog/scm/CMakeLists.txt) good but the cmake files don't need to be installed to datadir. Specially after deciding where to put the cmake files, say /use/local/lib/cmake/{cogutil,atomspace,opencog} it's best if it's consistent like mentioned in the comment above

noskill:
@Dagiopia Ok, that makes sense, i'll move cmake files to ${LIB_INSTALL_DIR}/cmake/CogUtil

PR #128, opened 2018-08-29
move cmake files to lib/cmake/CogUtil (fix for #126)

noskill:
Moving cmake files to ${LIB_INSTALL_DIR}/cmake/CogUtil for consistency with atomspace and opencog. See issue #126

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -43,9 +43,11 @@ INSTALL(FILES
 set(INCLUDE_INSTALL_DIR include/ )
 set(LIB_INSTALL_DIR lib/ )
 
+set(COGUTIL_CMAKE_DIR lib/cmake/CogUtil)
+
 include(CMakePackageConfigHelpers)
 configure_package_config_file(CogUtilConfig.cmake.in ${CMAKE_CURRENT_BINARY_DIR}/CogUtilConfig.cmake
-                              INSTALL_DESTINATION ${LIB_INSTALL_DIR}/CogUtil/cmake
+                              INSTALL_DESTINATION COGUTIL_CMAKE_DIR
                               PATH_VARS INCLUDE_INSTALL_DIR LIB_INSTALL_DIR)
 
 write_basic_package_version_file(
@@ -58,5 +60,5 @@ INSTALL(FILES
        ${CMAKE_CURRENT_BINARY_DIR}/CogUtilConfigVersion.cmake
        ${CMAKE_CURRENT_BINARY_DIR}/CogUtilConfig.cmake
        DESTINATION
-       ${LIB_INSTALL_DIR}/CogUtil/cmake)
+       ${COGUTIL_CMAKE_DIR})
 


cmake/CMakeLists.txt
```diff
@@ -38,14 +41,13 @@ INSTALL(FILES
         OpenCogMacros.cmake
         COPYING-CMAKE-SCRIPTS
         DESTINATION
-        ${DATADIR}/cmake/)
+        ${CMAKE_INSTALL_DIR}/cmake)
```
Dagiopia: /cmake not needed here... the path is suppose to be /..../cmake/CogUtil

noskill: Yes, but all opencog projects append cmake to the path they load from CogUtilConfig.cmake. It doesn't seem to cause any problem, so I decided it is not worth to update all the projects.

vsbogd:
Hi @Dagiopia, @noskill, I would like to discuss this PR and https://github.com/opencog/cogutil/issues/126 issue further before merging it. 

As far I see original issue was about moving ```*.cmake``` files like ```FindPthreads.cmake``` and ```AddCxxTest.cmake``` into ```${INSTALL_PREFIX}/lib/cmake``` folder. I am not sure it is correct. 

First of all I would like to pay your attention that there are two different kinds of ```*.cmake``` files. ```Find*.cmake``` and ```*Config.cmake```. ```Find*.cmake``` are files which are used by opencog to find 3rd party libraries. ```*Config.cmake``` are files which are exported by opencog itself to allow opencog clients find opencog libraries. CMake ```find_package``` command also handles these kind of .cmake files separately. ```Find*.cmake``` files are searched in ```CMAKE_MODULE_PATH``` paths. ```*Config.cmake``` files are searched in the following folders (See [find_package documentation](https://cmake.org/cmake/help/v3.0/command/find_package.html)):
```
<prefix>/(lib/<arch>|lib|share)/cmake/<name>*/          (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/                (U)
<prefix>/(lib/<arch>|lib|share)/<name>*/(cmake|CMake)/  (U)
```

At the moment this ```Find*.cmake``` files are placed to the ```${INSTALL_PREFIX}/shared``` folder because it is similar to Ubuntu ```cmake``` package layout. This package also installs number of ```Find*.cmake``` files and keeps them under ```share``` folder. ```*Config.cmake``` files are usually placed under ```${INSTALL_PREFIX}/lib``` directory probably because it is better to locate them near libraries themselves.

To summarize I think that current ```*.cmake``` files locations for OpenCog libraries are consistent. The only exception is that CogUtil ```*Config.cmake``` files are placed under ```{$INSTALL_PREFIX}/lib/CogUtil/cmake``` instead of ```{$INSTALL_PREFIX}/lib/cmake/CogUtil``` which is not consistent with other libraries. 

@Dagiopia, I don't understand which cross-compilation issue do you mention in https://github.com/opencog/cogutil/issues/126. Do you have actual example of setup which doesn't compiles?

linas:
@Dagiopia your concern about cross-compilation is exactly correct. I don't know what the full fix needs to be.  We are not the first to ever cross-compile, sot there must be some common (standard?) way that everyone solves this.

vsbogd:
@Dagiopia, I am not sure that using DATADIR to hardcode data path into application is a it is proper way. But I don't know how to do it properly. I think one of the following ways should work:

- Make application using one prefix and install it with another. For instance use ```DATADIR=/usr/loca/share``` for building and ```DATADIR=/home/user/my_root/usr/local/share``` for installation. Actually it will fail, because ```*Config.cmake``` will contain ```/usr/loca/share``` as COGUTIL_DATA_DIR and Atomspace will look for ```Find*.cmake``` files there.

-  Use chroot for cross compilation. In such case you can use any paths you need to build it as in target filesystem. Anyway if your architecture is different you probably need libraries which are compiled for target platform.

- Use additional variable like you described above. But I would use this variable in C++ code to hardcode the data path.

Dagiopia:
yes, that was actually why I was saying there would be no escaping this problem since once DATADIR is set and the source files are modified, it should also be used to install the data files to that location so that the binary finally can find them. 
I always use DESTDIR when making packages and a chroot with qemu static emulator when compiling for arm but I thought to point this out so that it would be more easier for someone else to do it by setting a few variables on the command line. 
@linas I looked into vim now and I think they use environment variables to find a configuration file and and a full path is not embedded in the binary. 
Anyway it seems it's time to close this and I see that the path is now `.../lib/cmake/CogUtil` which is good.

Dagiopia:
saw ~ and $HOME to find the local vimrc file with paths like `$HOME/.vimrc` and `~/vim/.vimrc` embedded in the binary. When launched they set vars like `$VIM` `$VIMRUNTIME` and `$MYVIMRC`

PR #129, opened 2018-09-10
Single-queue message writing

linas:
Allow the use of multiple loggers, with one message queue per filename.

This is intended to fix issues discussed in #127


ngeiswei:
Looks great! I don't feel I need to review it in details because I trust your code but for what I care it passes the unit test.

PR #133, opened 2018-10-05
Add as-moses in circleci

ngeiswei:


ngeiswei:
Oh, I intended to merge that to singnet, no big deal but it makes the git history of that PR slightly more convoluted that it needs to be, sorry.

PR #135, opened 2018-11-12
Counter Improvements

rTreutlein:
Added boost::arithmetic1/2 and allowed passing of a comparator to the underlying map
Added test for all the new functionality

### DIFF
--- a/opencog/util/Counter.h
+++ b/opencog/util/Counter.h
@@ -1,4 +1,4 @@
-/** Counter.h --- 
+/** Counter.h ---
  *
  * Copyright (C) 2011 OpenCog Foundation
  *
@@ -8,12 +8,12 @@
  * it under the terms of the GNU Affero General Public License v3 as
  * published by the Free Software Foundation and including the exceptions
  * at http://opencog.org/wiki/Licenses
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU Affero General Public License
  * along with this program; if not, write to:
  * Free Software Foundation, Inc.,
@@ -39,7 +39,7 @@ using boost::adaptors::map_values;
 
 //! Class that mimics python Counter container
 /**
- * This is basically a dictionary of key:count values. 
+ * This is basically a dictionary of key:count values.
  * Given following pseudocode:
  * @code
  * for word in ['red', 'blue', 'red', 'green', 'blue', 'blue']:
@@ -52,7 +52,7 @@ class Counter : public std::map<T, CT>,
     boost::addable<Counter<T, CT>>
 {
 protected:
-    /** @todo this will be replaced by C++11 constructor 
+    /** @todo this will be replaced by C++11 constructor
      * delegation instead of init
      */
     template<typename IT>
@@ -98,7 +98,7 @@ public:
         typename super::const_iterator it = this->find(key);
         return it == this->cend()? c : it->second;
     }
-    
+
     //! Return the total of all counted elements
     CT total_count() const
     {
@@ -132,7 +132,18 @@ public:
             this->operator[](v.first) += v.second;
         return *this;
     }
-    
+
+    Counter& operator-=(const Counter& other) {
+        for (const auto& v : other)
+            this->operator[](v.first) -= v.second;
+        return *this;
+    }
+
+    Counter& operator*(const CT& num) {
+        for (const auto& v : *this)
+            this->operator[](v.first) = v.second * num;
+        return *this;
+    }
     /// @todo add method to subtract, multiply, etc Counters, or
     /// scalar and Counter, etc...
 };
--- a/opencog/util/Counter.h
+++ b/opencog/util/Counter.h
@@ -106,7 +106,7 @@ public:
     //! Return the total of all counted elements
     CT total_count() const
     {
-        return boost::accumulate(*this | map_values, 0);
+        return boost::accumulate(*this | map_values, (CT)0);
     }
 
     //! Return the mode, the element that occurs most frequently
--- a/opencog/util/Counter.h
+++ b/opencog/util/Counter.h
@@ -47,12 +47,12 @@ using boost::adaptors::map_values;
  * @endcode
  * the counter will hold 'blue': 3, 'red': 2, 'green': 1
  */
-template<typename T, typename CT>
-class Counter : public std::map<T, CT>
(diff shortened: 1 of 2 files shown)


linas:
After a quick skim, it seems OK to me. I'll let @ngeiswei review and merge.

Two minor comments:

* Is there a way of avoiding boost? I've been burned by boost too many times, and have become quite suspicious of it. Just don't like it.

* Counter seems to be a multi-set. I like multisets (aka bags). Some important theory to keep in mind:

-- multisets are a lot like mathematical vectors, but they aren't actually mathematical vectors. Why not, you ask?  Math vectors live in Euclidean space, have dot-products (cosine products) that are rotationally invariant, etc. Multisets store frequency-counts and are implicitly normalized to sum_i p_i = 1 bu dividing by the total count. i.e. a probability. The shape of sum_i p_i = 1 is a convex polytope, and it is NOT a sphere; it is not sphere-symmetric. So rotations convert probabilities into not-a-probability, and cosine products of probabilities are ill-behaved.  I am starting to understand that large swaths of confusion in machine learning is due to assuming that multiset == vector. .. which is false, if you are using multisets for frequency counts and probability distributions.

And then the usual: with appropriate care, you can create mutlisets in the atomspace, using `MemberLink` and `NumberNode` .. its somewhat awkward and inefficient, due to the atospace being like-a-database.

One can create multisets as values, which *are* efficient. We should almost certainly have a `MultiSetValue`, that would be very similar to this pull req. Seems like a good idea, just at this moment.

On can take tensor products of multisets, which works almost exactly the same way that tensor products of vectors do. So, a "matrix" in linear algebra is the same thing as a 2-tensor.

The `matrix` code is attempt to slap a matrix API on the atomspace. It should be extended to 1-tensors (aka multisets) and 3-tensors, and etc. It should be maybe re-designed in C++?? and certainly ported to R so that agi-bio can use it. 

What I'm saying is that I like what you are doing here, but take care about a slippery-slope: We need nice, easy-to-use, fast, clear-cut API's, a lot like what you have here, but in the atomspace, and for Values, in general.

linas:
And perhaps this is your intent for DistriibutionalTV:
```
class DistributionalTV: pulbic TruthValue, Counter<something, double>
{ ...};
```

opencog/util/Counter.h
```diff
@@ -117,30 +119,111 @@ class Counter : public std::map<T, CT>,
         return key;
     }
 
+	/* Counter operators:
```
ngeiswei: @rTreutlein , here and elsewhere tabs are used for indentation, while space is the dominate indentation style in this file. 

Well [smart] tabs is the dominate indentation style overall, so alternatively you could convert the rest in tabs, but would be better kept for another PR. Anyway, choose what you prefer...

ngeiswei:
@linas said
> Is there a way of avoiding boost?

boost.operators is used a bit all over the place, not just in Counter. It looks like removing it enough so it doesn't pollute the atomspace code would be some moderate amount of work. Here are the places it is used:

cogutil
```
opencog/util/Counter.h
opencog/util/functional.h
```
atomspace
```
opencog/atoms/core/Variables.h
opencog/atoms/core/Context.h
opencog/atoms/core/Quotation.h
opencog/rule-engine/Rule.h
opencog/rule-engine/backwardchainer/BIT.h
opencog/rule-engine/forwardchainer/SourceSet.h
opencog/unify/Unify.h
```

Obviously I'm the main culprit for introducing it (heavily influenced by Moshe Looks) but I never had any problem with it, quite the contrary, I recall saving a lot of hassles (specifically with MOSES, when I learned it) by not having to overload a countless number of operators, reducing the chances of introducing bugs, etc.

I think boost.operators is quite stable, according to the revision history in the comment of operators.hpp it is pretty much fixed in stone. Can you point to a specific problem you had with it?

rTreutlein:
@linas Yes something like that atm it looks like this:
```
class DistributionalValue : public ProtoAtom {
Counter<IntervalList, double> value; 
...}
```
Where an IntervalList is a vector of a vector of doubles. I need the 2 vectors as an Interval can have 1 value to describe a singleton-set or 2 to describe a normal Interval. And for joint-distributions, we might have more than 1 interval.

Regarding your MultiSetValue that is similar to something i had at one point. The problem is just that having a Handle stored inside of a Value leads to some problems as described in https://github.com/opencog/atomspace/issues/1884 . Mainly that you can't create them without the atomspace. Which can be fine depending on the situation but for TruthValues a lot of the code creates them without access to the atomspace. So it is something I want to avoid for DVs. 
Instead, as you suggested I want to create an API that can fish the data from wherever that is stored (efficiently in a value or more flexible in the atomspace directly).

ngeiswei:
@linas said
> So rotations convert probabilities into not-a-probability, and cosine products of probabilities are ill-behaved. I am starting to understand that large swaths of confusion in machine learning is due to assuming that multiset == vector.

That's interesting, never considered that... Though I don't see where there is a temptation to rotate probabilities (and how it relates to vectors in that respect).

linas:
> boost

Not that big a deal. No need to strip it all out; simply caution is advised. The biggest boost offender was `boost::singals2` which was was ten times slower than the current homebrew signals used in the atomspace. I can sort-of-understand how the developer of signals2 ended up that way ... trying to be all things to all people, but the overhead was huge.  Other issues I remember only vaguely, something about threading, and some other issues with iterating.   Long time ago.

linas:
> MultiSetValue

@rTreutlein OK. I was trying to encourage you to think through the implications of as many design alternatives as you can, before settling on one. Create the design you need. ... Although InvertvalList sounds scary.  Don't over-engineer; if you can solve the problem with two or three special cases, that might be better than one does-everything class.

linas:
> rotate probabilities

Well, that is exactly what SVD does, and it's knock-offs -- latent semantic indexing (LSI), latent semantic analysis (LSA)  Whenever you hear the words "frobenius norm" or "nuclear norm" or "singular value", there were rotations applied implicitly or explicitly.  Its really quite bizarre how widespread this is, and when I look, it seems to come down to just one or two papers where the authors tried 3 or 4 different norms, got and accuracy of (something like this; I'm making these up from memory:) 42% 46% 57% and 59% with the 59% being cosine distance, so ah hah it must be the best! (In my mind, 57% and 59% is a "tie", given the noise, and given that no  error bars, estimates or error analysis was performed/published) If you read the original papers, its just kind of old and on some old small hokey dataset.  so wtf. And the rest of the world is like "oh, yeah, cosine distance, I remember that from college, lets use it!" This last sentence described me until I started looking at it more carefully...

opencog/util/Counter.h
ngeiswei: Just to be clear we're talking about smart/mixed tabs (tabs for indentation, space for alignment, see https://wiki.opencog.org/w/Development_standards#Formatting).

Taste is taste, but you might actually learn to prefer tabs, for one thing it makes moving around snappier.

PR #136, opened 2018-11-14
Spaces to Tabs for all .cc;.h;.cxxtest files

rTreutlein:
As I discussed with Nil I switched all the indentation to Tabs as this is the dominant style in the other repos.

Would be nice if somebody could take a closer look, as this change looks fine in my editor but on GitHub some things no longer seem aligned. for example Counter.h line 64.

### DIFF
--- a/opencog/util/Config.cc
+++ b/opencog/util/Config.cc
@@ -57,16 +57,16 @@ using namespace std;
 static char const* blank_chars = " \t\f\v\n\r";
 static string strip(string const& str, char const *strip_chars = blank_chars)
 {
-    string::size_type const first = str.find_first_not_of(strip_chars);
-    return (first == string::npos) ?
-        string() :
-        str.substr(first, str.find_last_not_of(strip_chars) - first + 1);
+	string::size_type const first = str.find_first_not_of(strip_chars);
+	return (first == string::npos) ?
+		string() :
+		str.substr(first, str.find_last_not_of(strip_chars) - first + 1);
 }
 
 
 Config* Config::createInstance()
 {
-    return new Config();
+	return new Config();
 }
 
 Config::~Config()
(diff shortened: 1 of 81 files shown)


ngeiswei:
Wow, I didn't necessarily mean to convert all files, only Counter, but, well, awesome, I'll take it.

I have set my editor to display 3 spaces for tabs, so I usually catch whitespace tabs pretty easily. I'll have a look at all that tomorrow (if @linas doesn't close that PR by then ;-)).

opencog/util/Counter.h
```diff
@@ -43,196 +43,196 @@ using boost::adaptors::map_values;
...
-    }
+	/** @todo this will be replaced by C++11 constructor
+	 * delegation instead of init
+	 */
+	template<typename IT>
+	void init(IT from, IT to) {
+		while(from != to) {
+			this->operator[](*from) += 1;  // we don't use ++ to put the
+										   // least assumption on on CT
+			++from;
```
linas: Counter.h line64: here's the problem: its mis-using the smart-space thing. It should be three tabs, followed by 32 spaces, followed by the slashes and comment.  Perhaps this seems awkward. I've concluded that post-code comments are awkward in general; whenever I see one of these, I put it on a line before the code. In this case, the whole comment can fit on one line, if its put on a line by itself, before the code.  That seems cleaner.

linas:
The problem with global whitespace changes like this is that it erases the file blame log.  I use blame logs rarely; when I do, here's how. Say, there's some bug. You find the bug in some file, look at the code, and either you just fix it, or maybe instead perhaps think "gee that looks weird. Who would have done something weird like that, and why?"  So you look at the blame log, to find the last person to modify that line of code, and what commit that was.  Then look to see what else that commit did.  If it looks normal, then you fix the bug and forget about it.  If the whole commit still looks wrong, then look at all the other commits that came in with the same pull request -- maybe the whole pull req is bad, and has to be undone.

If you merge this, then the blame log is effectively erased: the last person to touch most of the code will be this rTreutlin character. Clicking through, it becomes clear that he's not the one to blame, so now you need to look at some older version of the file to find the blame. It's just one more step, and more steps are always annoying.

Here, I kind-of don't care, because debugging cogutils is an unlikely activity, and then looking at the blamelog for it is even more unlikely. So I kind of don't care.  This can be merged.

There is one case to watch out for, and I've seen it happen in real life, more than once, and its quite evil.  Some junior developer makes a multi-hundred line-of-code whitespace change, and in that change, "secretly hidden away" is a non-trivial non-whitespace change. And its usually scandalous, because it usually introduces some bug. Since it now took some senior developer to find the bug, typically over the Christmas holidays, or some particularly beautiful sunny weekend, the senior developer shows up to work on Monday morning and yells at his manager, yells at the junior developer, and calls some executive to complain about the f'd up process that deprived him of Christmas/great weekend. And then the executive decides to "fix" the process, usually by making things worse for everyone.  I kid you not. This happens in big corporations. Disclaimer: none of the actors in this story was me; rather, this turns into lunchtime gossip.

Moral of the story: global pure-whitespace changes tend to get the evil eye.

opencog/util/Counter.h
ngeiswei: @linas is correct. @rTreutlein what tool did you use to format? Some smart-tab Neovim plugin?

ngeiswei:
I agree with all your comments, @linas, and this gossip should probably be turned into a biblical story.

I suppose the default should be that one can fix style one file at a time if that file contains a mixture of 2 styles. In that case however I think it's OK to go nuts, should have little or no repercussions, and I'll carefully review it to make sure God(el) doesn't punish us.

ngeiswei:
@rTreutlein , you've reused the same branch `HandleCounter`, that's only gonna add to the confusion. The branch name is stored in the git history, so it is of value to have a meaningful name, especially for such pervasive change.

If you could create a new PR with a new branch name that would be appreciated, meanwhile I'm having a look at it...

opencog/util/Config.cc
```diff
@@ -75,334 +75,334 @@ Config::~Config()
...
+	// Finish configuring the logger... The config file itself
+	// contains the location of the log file. This is working around
+	// a chicken-and-egg problem with reporting config file issues.
+	// Such is life; this is a lot easier than debugging screwed-up
+	// file-path craziness in a debugger. We MUST log the path!!!
+	setup_logger();
+
+	// And then finally, at long last!!! report what happened.
+	logger().info("Using config file found at: %s\n",
+				  path_where_found().c_str());
```
ngeiswei: Not properly aligned, maybe you haven't used the proper tool/plugin... It looks like the default plugin for smart-tabs in vim is https://www.vim.org/scripts/script.php?script_id=231 (no idea how that works with Neovim).

opencog/util/Config.cc
```diff
@@ -75,334 +75,334 @@ Config::~Config()
...
 
 const bool Config::has(const string &name) const
 {
-    if (_no_config_loaded)
-        logger().warn("No configuration file was loaded! Param=%s",
-                      name.c_str());
-    return (_table.find(name) != _table.end());
+	if (_no_config_loaded)
+		logger().warn("No configuration file was loaded! Param=%s",
+					  name.c_str());
```
ngeiswei: Not aligned.

opencog/util/Config.cc
```diff
@@ -75,334 +75,334 @@ Config::~Config()
...
+	return (_table.find(name) != _table.end());
 }
 
 void Config::set(const std::string &parameter_name,
-                 const std::string &parameter_value)
+				 const std::string &parameter_value)
```
ngeiswei: Not aligned.

opencog/util/Config.cc
```diff
@@ -75,334 +75,334 @@ Config::~Config()
...
 {
-    if (not has(name))
-       throw InvalidParamException(TRACE_INFO,
-                                   "[ERROR] parameter not found (%s)",
-                                   name.c_str());
-    return _table.find(name)->second;
+	if (not has(name))
+	   throw InvalidParamException(TRACE_INFO,
+								   "[ERROR] parameter not found (%s)",
+								   name.c_str());
```
ngeiswei: Not aligned.

opencog/util/Config.h
```diff
@@ -33,83 +33,83 @@
...
  * Nil: if overwrite is true then the static variable instance@n
- *      is changed with the createInstance provided@n
- *      it is a temporary dirty hack@n
+ *		is changed with the createInstance provided@n
+ *		it is a temporary dirty hack@n
  */
 typedef Config* ConfigFactory(void);
 Config& config(ConfigFactory* = Config::createInstance,
-               bool overwrite = false);
+			   bool overwrite = false);
```
ngeiswei: Not aligned.

opencog/util/Counter.h
```diff
@@ -43,196 +43,196 @@ using boost::adaptors::map_values;
  * Given following pseudocode:
  * @code
  * for word in ['red', 'blue', 'red', 'green', 'blue', 'blue']:
- *   cnt[word] += 1
+ *	 cnt[word] += 1
```
ngeiswei: Oh, I get how you've dealt with that! You've performed a "    " -> "\t" replacement. But that's not good enough

1. doesn't align
2. create this sort of wrongs

A lot worse could happen too, like changing a string value, etc, leading to the kind of archetypal stories St Linas has just recounted.

Can't you have Neovim apply smart-tabs to all files in one go? I'm sure you can, if not, please migrate to the other church... ;-)

ngeiswei: The other option is to do a regex replacement, using `^` to indicate the start of a line, etc. But I think asking your editor to smart-tabs the whole thing is better as it will properly be aligned.

ngeiswei: Also, maybe create one commit per file (or h/cc pair), just in case one may want to bisect some problem, very unlikely IMO, just suggesting... Or if you prefer, only take care of Counter, it's up to you, just follow you're personal aesthetic inclinations.

rTreutlein: Well, I just used one shell command to retab all files using neovim. So 1 commit per file would be a lot more effort. 
And as I said everything is still aligned in my editor. So I guess I'll try again using the plugin you suggested. If that doesn't help then I'll just do Counter and leave the rest for future changes.

PR #137, opened 2018-11-15
White space fix

rTreutlein:
@ngeiswei So I tried using the RetabIndent from the plugin you recommend but the result is the same. It converts 4 spaces into 1 tab which means as long as you display 1 tab as 4 spaces everything is still aligned correctly.
But since GitHub shows the first tab as 6 spaces long and all others as 8 spaces long this, of course, breaks the alignment. Which shows nicely why I prefer spaces.
Anyway, I fixed the misaligned end of line comment by moving it into its own line.

### DIFF
--- a/opencog/util/Counter.h
+++ b/opencog/util/Counter.h
@@ -49,190 +49,190 @@ using boost::adaptors::map_values;
  */
 template<typename T, typename CT, typename CMP = std::less<T>>
 class Counter : public std::map<T, CT, CMP>
-    , boost::arithmetic<Counter<T, CT, CMP>>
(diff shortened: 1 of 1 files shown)


ngeiswei:
@rTreutlein, multi-line comments are kinda special, they are no syntactic rules to know how to indent/align them. Have you tried on something 
```c++
	void init(IT from,
IT to)
```
It should insert one tab, then space till the second argument is aligned.

ngeiswei:
I would recommend that you do not render tabs with 4 spaces otherwise mixtures of spaces and tabs will go unnoticed.

PR #138, opened 2018-11-23
Synchronize version in git-tag with c++ macro 

AmeBel:


linas:
version 2.0.3 was tagged on Dec 12 2016 and there have been 215 commits since then. 

I think I made all four tags on that repo. I do not have any current process for determining when to bump up to a new version number.  I do not plan to create such a process, but perhaps its time to have one.

It appears that I failed to sign any of the tags.

PR #139, opened 2018-11-30
Move OpenCog* cmake files to atomspace

kasimebrahim:


kasimebrahim:
The install directory in atomspace/cmake is the same as cogutil/cmake. That is DATADIR/cmake so it should work fine with no modification.

kasimebrahim:
Or may be should I change the COGUTIL_DATA_DIR to DATA_DIR for consistency.

linas:
Ahhh. I see. The name COGUTIL_DATA_DIR is confusing, since its no longer part of cogutil, even if they happen to "accidentally" be the same install directory. It is very unlikely that COGUTIL_DATA_DIR and DATADIR will ever differ ... but it's still somehow confusing.

linas:
Per issue opencog/atomspace#1923 --I want to wait a few more days for the other repos update, before merging this one.

PR #143, opened 2018-12-10
ULP based float comarison

rTreutlein:
This function was and is still used in FloatValue and since it now is also being used in DistributionalValue I decided to move it into cogutils.


ngeiswei:
Interesting @rTreutlein.

The code in https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/ doesn't contain the initialization of `Float_t::i` but I suppose it is a pointer casting to int, as you did in `is_approx_eq_ulp`.

Remarks:
1. You've ignored the case where the floats have different signs, does it still behave well?
2. It seems tailored for double, not float or long double, right?

Maybe some unit tests (especially given the uncertainty of behaviors 1 and 2) would be welcome.

Does it work for big-endian (I'm especially when it's float or long double)? Not that we need to worry about unless we decide to port OpenCog to the Atari.

rTreutlein:
@ngeiswei I just reused the code that was in FloatValue which I think @linas wrote. So I can't really answer your questions directly, maybe linas can. What I can and will do is add those tests.

rTreutlein:
@ngeiswei Okay I have started writing the test and is_approx_eq_ulp does not work for floats only doubles haven't tried long double but I would guess that's also problematic. We could either limit this function to doubles it's all it's used for atm or maybe use something like boost::float_distance to calculate the ULPs.

Regarding different signs, I have found some inconsistent behaviour so I'll be adding an extra check for that.

ngeiswei:
`boost::float_distance` looks really cool, it might more expensive and @linas does like to rely on boost, but it is worth considering IMO.

linas:
max ulps for 32-bit floats is 23 or 24, I forget. max ulps for double is 52 I think, for long double its maybe 64 or 68. 

ulps is "unit of least precision", its how many bits there are in the mantissa of the float. Since floats are always denormalized, the compare is somewhat tricky. The code in FloatValue was .. a quick hack. Enough to get basic unit test cases to pass, not much more. Comparing floats is a tricky business.  Currently, the code is "almost unused"; some very old TruthValue unit tests need it, but that's it. Some of the old bugs on the space-server, time-server and allen-interval-algebra were about bad float-point compares .... 

Yes, I generally don't like boost, but a well-written, generic `float_distance` would be best.  There's no reason why it should be slow ... one complaint with boost is that they sometimes manage to convert a high-performance idea into a painfully slow implementation.

linas:
Ugh. I'm looking at `numeric.h` now. It seems to be a collection of outdated cruft. Stuff like `is_approx_eq` gives me the heebie jeebies; not at all obvious that this code is correct.  boost may well be the lesser of two evils.

The  `logsum` is not thread-safe,  ... but who is using this anyway? it appears to be the logarithm of the gamma function; great for number theory but unusual for any opencog-ish thing I can think of; I don't think it's in moses, either.   All the bithacks code -- surely that is cruft that no one uses? Lines 94-263 seem like candidates for removal.

rTreutlein:
Hm since the MaxULPs are different for each type the Value we compare against needs to depend on the type or at least needs to be lower. The 24 you originally picked workes fine for doubles but obviously can't work for floats.

And regarding the speed of boost::float_distance I could just run a simple performance test to see if there is any significant difference. I would guess not.

In regards to removing stuff, I noticed that the whole numericUTest is commented out in the CMakeLists.txt.

linas:
I'm pushing a pull req to remove cruft now, but I don't want to merge it because I'm not sure if it removes moses dependencies.

linas:
pul req #144

rTreutlein:
So, I have tested the speed of boost::float_distance and it's about 100 times slower than our basic version. I had to add checks for different signs but with that, I was not able to find a difference in the results of the 2 versions.

So i'll try to build my own version that also works with float and long double but might just be best to limit it to doubles. I mean I can't recall a place in OpenCog that uses something other than a double.

rTreutlein:
I am now thinking about if it makes sense to add another function that combines the ULP comparison with one using an absolute Epsilon. That would improve the problems ULP has when comparing with 0 and for FloatValues which might be 0 this seem like a good thing.

opencog/util/numeric.h
```diff
@@ -275,6 +276,25 @@ template<typename FloatT> bool is_between(FloatT x, FloatT min_, FloatT max_)
...
 
+// Compare floats with ULPS, because they are lexicographically
+// ordered. For technical explanation, see
+// http://www.cygnus-software.com/papers/comparingfloats/Comparing%20floating%20point%20numbers.htm
+bool is_approx_eq_ulp(double x,double y,long long int max_ulps)
+{
+	if ((x < 0) != (y < 0)) {
+		if (x == y) //incase x==0 y==-0
+			return true;
+		return false;
```
ngeiswei: I think you may just write `return x == y;`

ngeiswei:
@rTreutlein 
> I am now thinking about if it makes sense to add another function that combines the ULP comparison with one using an absolute Epsilon. That would improve the problems ULP has when comparing with 0 and for FloatValues which might be 0 this seem like a good thing.

That makes sense to me (though I'm not an expert in comparing floating point numbers).

linas:
> add another function that combines the ULP comparison with one using an absolute Epsilon

Seems plausible. I have no opinion. There is some likelyhood that one of the "how to compare floats" blog posts discusses this, and tells you that its a good idea, or a bad idea. (You did see the blog post? It's linked in the code, somewhere. It's long and requires thought. Many years ago, I understood it; now, I've mostly forgotten.)  I get nervous when some blog says "this is a bad idea".

ngeiswei:
IIRC, during my PhD, my code was doing a lot of numerical integration and I noticed that things were a bit more stable with long double, not dramatically better but still. The catch was that it was a bit slower.

All that to say that I can envision a day we may want to use long double.

rTreutlein:
Yea the article nil linked about talks about when to use what comparison. In short when comapring with 0 use an absolute methode, when comparing with something other then zero use a relative epsilone or ULP and when you don't know use a mix.

Althoug it seems like the is_approx_eq methode is already using such a mixed style so maybe we should use that in FloatValue? Anyway i think i'll leave things as they are right now. And if we want to continue discussing this it think we should move that discussion into a Issue.

linas:
long double in current versions of gcc is 80 bits, instead of 64 bits. So it uses twice as much space, for not that much more gain.  If it was 128 bits, that would be cool, but it's not. my attitude is that if you need more than double, you probably need a lot more than double.  The exception is .. numerical integration, I guess, where maybe the stability from long double allows for faster convergence, bigger steps, or something.  But we're not really doing either ...

PR #146, opened 2018-12-12
Minor Improvment to is_approx_eq_ulp

rTreutlein:
Forgot this small fix in the last PR


Dagiopia:
semicolon on line 286

rTreutlein:
fixed in #147

linas:
Can you PLEASE run the unit tests before merging? And make sure that atomspace builds? This has an obvious break that prevents other code from compiling:  Both of the added functions need to be declared `static inline`. I will fix this in just a few moments. ...

linas:
It's also got compiler warnings!

rTreutlein:
Yea sorry I thought I couldn't mess up such a small change which is stupid.

About the warnings, they were already in the original code so I thought it wasn't that important to fix. And it would make the code a bit more complex so I decided to just leave it as is.

PR #157, opened 2018-12-20
Install cogutil lib under PREFIX/lib/opencog intead of PREDIX/lib

### LINKED ISSUE #140: libcogutil.so installed in incorrect location.
It appears that libcogutil is being installed into `/usr/local/lib/libcogutil.so` instead of `/usr/local/lib/opencog/libcogutil.so` and this might (??) explain some of the weird behavior in various unit tests (e.g. failure to deal with the logger correctly).

Suggest that it should get installed into  `/usr/local/lib/opencog`, where everything else goes.

ngeiswei:
Fix #140


ngeiswei:
`moses` consistently fails on circleci but consistently passes on my systems. Can anybody else test that?

vsbogd:
@ngeiswei , I have tried it in docker, as far as I see in `moses` cmake uses python 2.7 and nosetests with python 3. This is the reason why test fails.

linas:
So, the policy question becomes: do we merge anyway because we know "it works"? Or do we wait for a moses fix?  (Since moses-python works with the old location, the fix for moses-python is to tell moses about the new location)

vsbogd:
I see that I was not precise in my conclusion. Actually I am not sure that `moses` finds `python2`. What I saw is that test fails because it uses `python3-nosetest` and test itself is written using `python2` syntax. I believe it is consequence of turning python unit tests on in `opencog-deps`. So may be solution is port python test to python3.

linas:
@vsbogd yes, everything needs to move to python3. But that does not explain why circleci pases with the old cogutils path. For that, I am attempting this: opencog/moses#77

linas:
Ohh.. that's interesting ..  so, circleci passed here opencog/moses#74 but now it fails .. so indeed, @vsbogd, it seems that my explanation doesn't explain it; I suppose you're right..

ngeiswei:
Thanks guys, I'm merging since it's clear the problem doesn't come from that PR.

linas:
I'm fixing the moses nosetests thing now.

PR #158, opened 2018-12-20
Initialize LogWriter::printToStdout to false

ngeiswei:
Otherwise the logger can randomly be initialized to output to stdout


vsbogd:
looks good to me

ngeiswei:
`moses` failed again, so I guess the problem is on moses' end, I'm merging.

PR #159, opened 2019-01-09
Add GCC options include file for cmake

vsbogd:
Add cmake include file which contains GCC options to be included in root CMakeLists.txt in other opencog repositories. Content of OpenCogGccOptions.cmake is copied from atomspace repo, see https://github.com/opencog/atomspace/blob/387588a33868accddf192bf4f051e8f54872184d/CMakeLists.txt#L51-L121


ngeiswei:
With little confidence on my part I would say that it looks OK. I'm merging based on trust.

PR #163, opened 2019-02-25
Fix Counter::operator*=(const Counter&)

ngeiswei:
Take into account zeros, i.e. values that are in one but not in the
other counter.

### DIFF
--- a/opencog/util/Counter.h
+++ b/opencog/util/Counter.h
@@ -150,15 +150,32 @@ public:
 		return *this;
 	}
 
-	//! multiply 2 counters,
+	//! multiply (inner product of) 2 counters,
 	/**
 	 * c1 *= c2
 	 * =>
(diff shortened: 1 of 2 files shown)


ngeiswei:
@rTreutlein, I'm confidence it is what we want, but please review as it might affect your code.

ngeiswei:
Also, it could be slightly faster by ignoring zeros from the other counter, which works semantically because defaults are zeros, however it would make the operator asymmetric at the data representation level, though still symmetric semantically.

I took the cautious approach and went full symmetric, but feel free to modify that if you feel it's better otherwise.

rTreutlein:
Yea that seems correct. And if I am not mistaken I don't actually use that particular operator so merging it is fine. In the unexpected case that this breaks something I would guess I have some other bug in the code.

So Merging this is okay.

PR #165, opened 2019-03-01
Fix the exception API

linas:
It was dereferenceing null pointers and other garbage!


linas:
Merging. BackwardChainerUTest is failing, but not due to this change.

PR #166, opened 2019-03-12
Wrap libbfd calls by mutex as libbfd is not thread safe

vsbogd:
Fast fix for issue https://github.com/opencog/atomspace/issues/2127


vsbogd:
Merging it as BackwardChainerUTest constantly fails on this CI

PR #168, opened 2019-03-19
Replace build status by CircleCI one

vsbogd:


linas:
Thanks.

Any clue if we could have distinct status for cogutil, atomspace and opencog? e.g. at this particular instant, atomspace is failing, because of a spurious double-free bug.  Its random, and unrelated to cogutil...

vsbogd:
I think right way is to debug this issue finally to understand why results are different for `cogutil` and `atomspace`. I have made several experiments and thought that it can be related to CircleCI caching and [this commit](https://github.com/opencog/cogutil/commit/04114f79437db429b53b6a2ada0c7a682d8bd149). Removing this commit fixed issue in my private branch, but surprisingly it didn't work in PR. Also it could be related to https://github.com/opencog/atomspace/issues/2130, looks like it is green right now because of PR https://github.com/opencog/atomspace/pull/2132.

PR #171, opened 2019-04-15
Fix set comprehension

ngeiswei:
C++ vector list and set comprehension is back!!! I had partially disable it due to a gcc bug and the bug has been fixed since then.

### DIFF
--- a/opencog/util/comprehension.h
+++ b/opencog/util/comprehension.h
@@ -114,44 +114,29 @@ auto list_comp(const Container& c, const Function& func,
     return l;
 }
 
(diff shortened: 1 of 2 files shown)


linas:
are we testing for the min required gcc to avoid this bug?  I guess gcc > 4.7?

ngeiswei:
The current minimum requirement is gcc 5.0, so yes someone (me) should test that. Thanks for the reality check.

ngeiswei:
It works for gcc 6.3.0.

ngeiswei:
It works for gcc 5.4.0 as well.

PR #180, opened 2019-07-10
Make build jobs less for moses

stellarspot:
Make build jobs less for moses to avoid "c++: internal compiler error: Killed (program cc1plus)" message


ngeiswei:
Good finding. That probably means that gcc runs out of memory on circleci.

AmeBel:
Even though the pull-request was a draft pull-request, I am merging as all tests are passing. Thanks :-)

PR #183, opened 2019-07-12
Cache GHC

ngeiswei:


ngeiswei:
Again failing because `The specified paths did not match any files in /ws`, feels it's a bug on circleci's end.

PR #186, opened 2019-07-24
Use circleci checkout comment instead of git clone

ngeiswei:
Otherwise it checks out the master not the current branch!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,9 +16,7 @@ jobs:
           keys:
             - ccache-{{ checksum "/tmp/date" }}
             - ccache-
-      - run:
-          name: Checkout CogUtil
-          command: git clone --depth 1 https://github.com/$CIRCLE_PROJECT_USERNAME/cogutil .
+      - checkout
       - run:
           name: CMake Configure
           command: mkdir build && cd build && cmake ..


ngeiswei:
Using `checkout` take 1m, with the following message
```
Either git or ssh (required by git to clone through SSH) is not installed in the image. Falling back to CircleCI's native git client but the behavior may be different from official git. If this is an issue, please use an image that has official git and ssh installed.
Enumerating objects: 44, done.
Counting objects: 100% (44/44), done.
Compressing objects: 100% (21/21), done.
Total 141166 (delta 20), reused 36 (delta 13), pack-reused 141122
Enumerating objects: 163, done.
Counting objects: 100% (163/163), done.
Compressing objects: 100% (131/131), done.
Total 240 (delta 81), reused 33 (delta 31), pack-reused 77
```
which is what prompt me to use `git clone --depth 1` which was obviously very dumb.

So I guess I'm merging that now. Unless perhaps someone more familiar with circleci than myself wants to review it?

ngeiswei:
Looks like a bug has been introduced in the atomspace that breaks the unifier, which of course was not caught by circleci due to another bug introduced by a moron I won't name. ;-)

I'm gonna look into that bug ASAP, meanwhile I think it's still good to merge that PR, better have broken test than broken tester.

vsbogd:
@ngeiswei , looks good to me.

Regarding 1m time I believe reason can be that by default history is checked out. Probably it can be fixed using custom clone command with CircleCI variable which refers to commit under test.

PR #187, opened 2019-08-31
Normalize check for /etc/os-release for other distributions that lack it

xloem:
I'm on Android using Termux, atm, where this path doesn't exist.


linas:
Thanks

PR #188, opened 2019-08-31
Add specific check for unix backtrace

xloem:
Aids building on other platforms than just apple and cygwin, that lack this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -129,6 +129,15 @@ IF (NOT CXXTEST_FOUND)
 ENDIF (NOT CXXTEST_FOUND)
 
 # -------------------------------------------------------------
+# Look for GNU C++ backtrace. Needed for basic stack printing.
+FIND_PACKAGE(GNUBacktrace)
+IF (GNU_BACKTRACE_FOUND)
+	ADD_DEFINITIONS(-DHAVE_GNU_BACKTRACE)
+ELSE (GNU_BACKTRACE_FOUND)
+	MESSAGE(STATUS "GNU backtrace missing: No basic stack-trace printing.")
+ENDIF (GNU_BACKTRACE_FOUND)
+
+# -------------------------------------------------------------
 # Look for binutils-dev. Needed for pretty stack printing.
 FIND_PACKAGE(BFD)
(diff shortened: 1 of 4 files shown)


linas:
This is confusing and incorrect -- one has to have `libbfd` and also `libiberty` and also `execinfo.h` to have backtraces. The check for `execinfo.h` should not be needed, since it comes automatically with `libc-dev` and if you don't have `libc`, far ore is likely to be broken.  Is your libc actually missing `execinfo.h` ??

I don't mind the general intent here -- to create a `UnixBacktrace.cmake`, but that file should then contain all the right stuff i it.

linas:
I'm going to close for now, please re-open when you've reorganized everything...

xloem:
hi, uclibc-dev and non-linux platforms often don't have execinfo.h, creating issues compiling the backtrace code.  this was to aid building in Termux on Android.  I'll do it properly and resubmit if I set up that environment again.

linas:
Yes, please. Thank you for the work, it just looked half-finished, and sometimes merging half-finished code makes things worse, not better.

Are you thinking of porting the atomspace as well?

xloem:
I think we had a miscommunication.  by "unix backtrace" I simply meant the availability of `backtrace(3)` and `backtrace_symbols(3)`, which are used by the current backtrace code even if `libbfd` and `libiberty` are unavailable.  These gnu linux extensions are only present if execinfo.h is present.  Would you accept a cosmetic change to the code to make these clearer, such as renaming `FindUnixBacktrace.cmake` to `FindGNUExecInfo.cmake`?

Yes, I have made some progress on porting atomspace as well.

linas:
Ah, OK, yes, indeed, my fault, I misunderstood; it's been so long that I forgot that the backtrace itself was distinct from the pretty-printing.  Yes, `FindGNUExecInfo` would indeed help lessen the confusion!  Thanks!

linas:
I re-opened the pull req, but now there are conflicts in cmake ...

xloem:
I force-pushed a different changeset.  How's this look?

linas:
Thank you, this looks good.

PR #189, opened 2019-08-31
Use platform-independent float ulp type

xloem:
Using only `int64_t` in this check lets it work properly on platforms with differing sizes for `long int` or `long long int`.  As of C++11, `std::abs` has overloads that will pick the proper implementation.


linas:
thanks

PR #191, opened 2019-09-01
clang-compatible fgetc return type

xloem:
Clang assumes `char == -1` is always false, so this change to respect the `fgetc` function signature better allows the loop to complete and the test to pass on my system.


linas:
yes, thanks!

PR #192, opened 2019-09-01
respect TMPDIR in LoggerUTest

xloem:
Required for test to pass when `/tmp` is not writable.

### DIFF
--- a/tests/util/LoggerUTest.cxxtest
+++ b/tests/util/LoggerUTest.cxxtest
@@ -27,6 +27,7 @@
 #include <unistd.h>
 
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
 #include <fstream>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


linas:
Interesting. When would `/tmp` never be writable? How could that ever happen?

xloem:
Sorry for not including the context.  This is true on Android.  I was building on Termux, which runs as an Android app.

linas:
Huh, OK, and one more sanity-check question: Android always has the `TMPDIR` environment variable set?

xloem:
That's true for me, but I don't think it's guaranteed (on Android or in general, hence the default specified at https://github.com/opencog/cogutil/pull/192/commits/31f6cf9081eb91d400ca6338a055d24d17882137#diff-8420f6067ca80fe4c0eeadd54d3093d0R81 ).  Termux is an app that provides a debian-like environment on Android for developer work, and sets `TMPDIR` by default.  The default Android shell on my phone also sets `TMPDIR`, but is very simplified and is usually not used to build apps like opencog.

The value on my Android is `/data/local/tmp`.  Termux changes this to point inside a sandbox in its local storage (`/data/data/com.termux/files/usr/tmp` for me).

Thanks for checking my work.  I make many mistakes.

linas:
OK, that seems reasonable.

linas:
thanks for the pull req. I'm just unfamiliar with that, and had to understand what it meant.

linas:
BTW, Are you working on a 32-bit or a 64-bit CPU?  Do you know which kernel is being used? Is it `armhf`? There is an unrelated bug for another opencog component opencog/link-grammar#1000 which is failing a 32-bit armhf build on Ubuntu. (There are no cogutils builds for debian, as far as I know; no package maintainers)

linas:
Actually, @xloem  if it's not too hard, could I ask you to test link-grammar on your system? Just curious ...

Issue #195, opened 2019-09-12
Wrong interaction between stdout flag of different loggers


ngeiswei:
## Problem

Setting the stdout flag of a logger consequently, and incorrectly, sets the stdout flags of other loggers to that same value.

## Reproducibility

Enable `LoggerUTest::xtestLoggerStdoutFlagInteraction` (i.e. remove `x`)

https://github.com/opencog/cogutil/blob/64bbb63f0aa5b061ded1a42a1040bae2cb25d159/tests/util/LoggerUTest.cxxtest#L287-L298

PR #198, opened 2019-10-07
Move printToStdout out of the log writer

ngeiswei:
So that each logger can set this flag independently of the other
loggers, regardless of whether they share the same log file.

Fix for issue https://github.com/opencog/cogutil/issues/195


ngeiswei:
OpenPsi is failing, however it was failing before as well, so I'm merging.

opencog/util/Logger.cc
```diff
@@ -262,13 +261,6 @@ void Logger::LogWriter::write_msg(const std::string &msg)
 
     // Stdout writing must be unlocked.
     lock.unlock();
```
linas: this comment is now orphaned. And the unlock happens automatically.

PR #203, opened 2019-10-14
Circle ci attention test 2

vsbogd:


.circleci/config.yml
```diff
@@ -142,9 +142,6 @@ jobs:
       - run:
           name: Run tests
           command: cd build && make -j2 test
-      - run:
-          name: Install URE
```
ngeiswei: The name is wrong, yes, but you still do want to install cogserver and attention, right?

ngeiswei: I've fixed the names in https://github.com/opencog/cogutil/pull/201/commits/1f1958458114ad9377027f3170d4281d86e3df02 BTW, but it wouldn't expect to have any effect.

vsbogd: Ok :-) In fact my way of thinking was installing is not needed because you will install this component in next jobs. For example it is not required installing `attention` here because it is installed in `opencog` related job.

vsbogd: After your comment I think it can be useful to install it here to check that installation step works.
I am not sure about effect as well as I wrote in https://github.com/opencog/cogutil/pull/201#discussion_r333910527, but I just would like to check whether image size will be different for `attention` and `ure`.

vsbogd: Finally I am wrong about this step affects the build results: https://circleci.com/gh/opencog/cogutil/739?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

Issue #176, opened 2019-06-12
Refactor get_python_lib from AtomSpace/OpenCog into CMAKE macro


ferrouswheel:
Both AtomSpace and OpenCog have a duplicate of `scripts/get_python_lib.py` for finding the correct python install path, introduced by opencog/atomspace#2220 and opencog/opencog#3535

This should be rewritten as a Cmake macro and added to cogutil's FindPython.

Per request here: https://github.com/opencog/opencog/pull/3535#issuecomment-501103412

PR #211, opened 2019-11-28
Fix atomspace cross compilation

vsbogd:
For cross compilation builds one need to set `PYTHON_DEST` manually as it cannot be determined during build using target Python interpreter. Goal of this PR is to move Python initialization code into single place and allow user overriding `PYTHON_DEST` by setting `PYTHON_INSTALL_PREFIX` parameter.


vsbogd:
#176  is also related

vsbogd:
@noskill please take a look

cmake/OpenCogFindPython.cmake
```diff
@@ -0,0 +1,90 @@
...
+     (2.7.0 VERSION_LESS ${PYTHONLIBS_VERSION_STRING})))
+	SET (HAVE_PY_LIBS 1)
+	MESSAGE(STATUS "Python ${PYTHONLIBS_VERSION_STRING} libraries found.")
+ELSE()
+	MESSAGE(STATUS "Python libraries NOT found.")
+ENDIF()
+
+# Cython is used to generate python bindings.
+IF(HAVE_PY_INTERP)
+	FIND_PACKAGE(Cython 0.23.0)
```
noskill: maybe we can use any cython newer than 0.23?

vsbogd: Excuse me didn't get. Actually this `FIND_PACKAGE(Cython 0.23.0)` should find any Cython which is not older than 0.23. `EXACT` is an option which forbids using other versions and there is no `EXACT`.

dagims:
@vsbogd it works well. Finds the python interpreter from the build system and the lib from host.
I was wondering about cogutil though because CogUtilConfig.cmake sets a static path to its installation location and find_package(CogUtil) ends up with COGUTIL_INCLUDE_DIR either set to /usr/local/include or <host_fs_path>/usr/local/include which in either case isn't desirable because the former would disable including the cogutil headers and linkining against the lib and the latter would be an issue when compiling something on the host system that depends on cogutil by providing cmake with a nonexistant path.

can we add a prefix to the paths in CogUtilConfig.cmake such as COGUTIL_PATH_PREFIX which should be set by the user (on command line or a toolchain file) to prefix the path with the path to the files of the host system?
for example:
```
set(COGUTIL_INCLUDE_DIR "${COGUTIL_PATH_PREFIX}/usr/local/include")
set(COGUTIL_DATA_DIR   "${COGUTIL_PATH_PREFIX}/usr/local/share/opencog")
set(COGUTIL_LIBRARY "${COGUTIL_PATH_PREFIX}/usr/local/lib/opencog/libcogutil.so")
set(COGUTIL_FOUND 1)
```
if the var is not set the path will just be /usr/local/include etc...

vsbogd:
@dagims, when you are talking about "host" you are meaning the system which will run atomspace after compilation, aren't you? I used word target to designate the same.

I see the problem you described. Yes it is impossible to set variables like `COGUTIL_INCLUDE_DIR` at compilation stage because they should have two different values for cross compilation and compilation on target system. Anyway it is not clear to me how to add such prefix variable.

For example in my build I set `CMAKE_INSTALL_PREFIX` ("/home/target/opt/opencog") and it includes both:
- path to the target filesystem on build machine ("/home/target");
- prefix from the root of target filesystem ("/opt/opencog").

After installation if I use cogutil from build machine I would like to keep both parts of the `CMAKE_INSTALL_PREFIX` ("/home/target/opt/opencog"). If I use cogutil from target machine I would like to replace "/home/target" by empty string. But during compilation cmake has only full `CMAKE_INSTALL_PREFIX` and doesn't know how to get "opt/opencog" from it.

vsbogd:
added commit to fix https://github.com/opencog/cogutil/issues/176

vsbogd:
> But during compilation cmake has only full CMAKE_INSTALL_PREFIX and doesn't know how to get "opt/opencog" from it.

Looks like [CMAKE_STAGING_PREFIX](https://cmake.org/cmake/help/latest/variable/CMAKE_STAGING_PREFIX.html) can help.

linas:
I typed in "cmake cross compilation" into google, and got this page: https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html  and there seem to be blogs about this as well. Differing targets for build-vs-install for cross-compilation is a common problem; someone must have written it up already, somewhere.

vsbogd:
Yes I studied https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html and couple of other information sources. Unfortunately I didn't find clear and precise recipe for this specific case. So experiments are still required to understand how it works.

I am merging it as this seems to be useful on its own. I will experiment and raise another PR on problem that @dagims spotted.

vsbogd:
@dagims are you ok with this?

PR #213, opened 2019-12-12
Implement attractive fix-point

ngeiswei:
For the definition of fix-point see https://en.wikipedia.org/wiki/Fixed_point_(mathematics)

### DIFF
--- a/opencog/util/algorithm.h
+++ b/opencog/util/algorithm.h
@@ -402,6 +402,24 @@ void clear_by_swap(C& c)
 	c.swap(empty);
 }
 
+/**
+ * Return the attractive fixpoint of a unary function and its
+ * argument, if it exists.  If no attractive fixpoint exists then it
+ * may never return.
+ *
+ * For the definition of fixpoint see
+ *
+ * https://en.wikipedia.org/wiki/Fixed_point_(mathematics)
+ */
+template<typename F, typename A>
+A fixpoint(const F& fun, const A& arg)
+{
+	A res = fun(arg);
+	if (res == arg)
+		return res;
+	return fixpoint(fun, res);
+}
+
 /** @}*/
 } //~namespace opencog
 
--- a/tests/util/algorithmUTest.cxxtest
+++ b/tests/util/algorithmUTest.cxxtest
@@ -21,6 +21,7 @@
  */
 
 #include <opencog/util/algorithm.h>
+#include <cmath>
 
 using namespace opencog;
 using namespace std;
@@ -28,7 +29,8 @@ using namespace std;
 class algorithmUTest : public CxxTest::TestSuite
 {
 public:
-	void test_powerset() {
+	void test_powerset()
+	{
 		set<int> s = {1, 3, 5};
 		set<set<int> > ps = powerset(s);
 		set<set<int> > expect_set = { {}, {1}, {3}, {5},
@@ -38,7 +40,8 @@ public:
 		TS_ASSERT_EQUALS(ps, expect_set);
 	}
 
-	void test_powerset_exact() {
+	void test_powerset_exact()
+	{
 		set<int> s = {1, 3, 5};
 		set<set<int> > ps = powerset(s, 2, true);
 		set<set<int> > expect_set = { {1, 3}, {1, 5}, {3, 5} };
@@ -46,7 +49,8 @@ public:
 		TS_ASSERT_EQUALS(ps, expect_set);
 	}
 
-	void test_cartesian_product() {
+	void test_cartesian_product()
+	{
 		set<int> c = {1, 2, 3};
 		set<vector<int>> result = cartesian_product(c, 2);
 		set<vector<int>> expect = { {1, 1}, {1, 2}, {1, 3},
@@ -54,5 +58,11 @@ public:
 		                            {3, 1}, {3, 2}, {3, 3} };
 
 		TS_ASSERT_EQUALS(result, expect);
+	}
+
+	void test_fixpoint()
+	{
+		float fp = fixpoint(sqrtf, 2.0f);
+		TS_ASSERT_EQUALS(fp, 1.0f);
 	}
 };


opencog/util/algorithm.h
```diff
@@ -402,6 +402,24 @@ void clear_by_swap(C& c)
...
+ * may never return.
+ *
+ * For the definition of fixpoint see
+ *
+ * https://en.wikipedia.org/wiki/Fixed_point_(mathematics)
+ */
+template<typename F, typename A>
+A fixpoint(const F& fun, const A& arg)
+{
+	A res = fun(arg);
```
vsbogd: Nil, may be use loop instead of recursion to no overflow stack?
```c++
	A argc = arg;
	while (1) {
		A res = fun(argc);
		if (res == argc)
			return res;
		argc = res;
	}
```

ngeiswei: I suppose you are write in principle. I was hoping the compiler would be smart enough to avoid that but it isn't, see the assembly code below (Debug version, but the Release version is no better, just less readable):

```asm
template<typename F, typename A>
A fixpoint(const F& fun, const A& arg)
   2acfd:	55                   	push   rbp
   2acfe:	48 89 e5             	mov    rbp,rsp
   2ad01:	48 83 ec 30          	sub    rsp,0x30
   2ad05:	48 89 7d e8          	mov    QWORD PTR [rbp-0x18],rdi
   2ad09:	48 89 75 e0          	mov    QWORD PTR [rbp-0x20],rsi
{
	A res = fun(arg);
   2ad0d:	48 8b 45 e0          	mov    rax,QWORD PTR [rbp-0x20]
   2ad11:	f3 0f 10 00          	movss  xmm0,DWORD PTR [rax]
   2ad15:	48 8b 45 e8          	mov    rax,QWORD PTR [rbp-0x18]
   2ad19:	ff d0                	call   rax
   2ad1b:	66 0f 7e c0          	movd   eax,xmm0
   2ad1f:	89 45 fc             	mov    DWORD PTR [rbp-0x4],eax
	if (res == arg)
   2ad22:	48 8b 45 e0          	mov    rax,QWORD PTR [rbp-0x20]
   2ad26:	f3 0f 10 00          	movss  xmm0,DWORD PTR [rax]
   2ad2a:	f3 0f 10 4d fc       	movss  xmm1,DWORD PTR [rbp-0x4]
   2ad2f:	0f 2e c1             	ucomiss xmm0,xmm1
   2ad32:	7a 0a                	jp     2ad3e <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)+0x41>
   2ad34:	0f 2e c1             	ucomiss xmm0,xmm1
   2ad37:	75 05                	jne    2ad3e <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)+0x41>
		return res;
   2ad39:	8b 45 fc             	mov    eax,DWORD PTR [rbp-0x4]
   2ad3c:	eb 18                	jmp    2ad56 <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)+0x59>
	return fixpoint(fun, res);
   2ad3e:	48 8d 55 fc          	lea    rdx,[rbp-0x4]
   2ad42:	48 8b 45 e8          	mov    rax,QWORD PTR [rbp-0x18]
   2ad46:	48 89 d6             	mov    rsi,rdx
   2ad49:	48 89 c7             	mov    rdi,rax
   2ad4c:	e8 ac ff ff ff       	call   2acfd <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)>
   2ad51:	66 0f 7e c0          	movd   eax,xmm0
   2ad55:	90                   	nop
   2ad56:	89 45 dc             	mov    DWORD PTR [rbp-0x24],eax
   2ad59:	f3 0f 10 45 dc       	movss  xmm0,DWORD PTR [rbp-0x24]
}
   2ad5e:	c9                   	leave  
   2ad5f:	c3                   	ret    
```

ngeiswei: Hmm, yet it's a tail recursion, why the compiler can't deal with it?!

ngeiswei: This is the Release version

```asm
template<typename F, typename A>
A fixpoint(const F& fun, const A& arg)
   146d0:	55                   	push   rbp
   146d1:	48 89 fd             	mov    rbp,rdi
   146d4:	53                   	push   rbx
   146d5:	48 89 f3             	mov    rbx,rsi
   146d8:	48 83 ec 18          	sub    rsp,0x18
{
	A res = fun(arg);
   146dc:	f3 0f 10 06          	movss  xmm0,DWORD PTR [rsi]
   146e0:	ff d7                	call   rdi
	if (res == arg)
   146e2:	0f 2e 03             	ucomiss xmm0,DWORD PTR [rbx]
	A res = fun(arg);
   146e5:	f3 0f 11 44 24 0c    	movss  DWORD PTR [rsp+0xc],xmm0
	if (res == arg)
   146eb:	7a 13                	jp     14700 <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)+0x30>
   146ed:	75 11                	jne    14700 <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)+0x30>
		return res;
	return fixpoint(fun, res);
}
   146ef:	48 83 c4 18          	add    rsp,0x18
   146f3:	5b                   	pop    rbx
   146f4:	5d                   	pop    rbp
   146f5:	c3                   	ret    
   146f6:	66 2e 0f 1f 84 00 00 00 00 00 	nop    WORD PTR cs:[rax+rax*1+0x0]
	return fixpoint(fun, res);
   14700:	48 8d 74 24 0c       	lea    rsi,[rsp+0xc]
   14705:	48 89 ef             	mov    rdi,rbp
   14708:	e8 c3 ff ff ff       	call   146d0 <float opencog::fixpoint<float (float), float>(float ( const&)(float), float const&)>
}
   1470d:	48 83 c4 18          	add    rsp,0x18
   14711:	5b                   	pop    rbx
   14712:	5d                   	pop    rbp
   14713:	c3                   	ret    
   14714:	66 2e 0f 1f 84 00 00 00 00 00 	nop    WORD PTR cs:[rax+rax*1+0x0]
   1471e:	66 90                	xchg   ax,ax
```

I'm not enough proficient in assembly to understand what it's doing, but I can see is that there's a recursive call at 14708.

vsbogd: Looks like the reason is passing argument as reference which can lead to side effect. if I remove it it optimizes well:
```c++
template<typename F, typename A>
A fixpoint(const F& fun, const A arg)
```

ngeiswei: Ah, good to know. Still weird that it cannot optimize it when passed by ref though.

vsbogd: > Still weird that it cannot optimize it when passed by ref though.

Looks like restriction of const semantics: compiler cannot guarantee that object state is not changed even when it is passed by const reference.

ngeiswei: Yeah, I understand, one can use `mutable` and so `const` cannot be a garanty, but I still don't understand why it requires by copy argument passing in `fixpoint` to optimize it. Seems to me the optimization still works even when passed by ref, but maybe I'm missing something.

linas:
I don't really like this, for several reasons. One is the compare 'if (res == argc)` which is faulty for floating point: many functions that "should" have a fixed point might, in an actual implementation bounce between two numbers differing in the least-significant bit. This can happen both for specific coding concerns, but also due to IEEE754 but also famously because 0.99999... = 1.00000... so really an epsilon is needed. 

But floating-point epsilons are notoriously difficult, as revealed by many blogs, at least half of which are wrong.

Also, its not hard for the user to write a simple loop, as needed.  Simpler to write a loop, than to hunt down this function in cogutils.

And finally, a fixed point is just a "periodic point of period one", and bouncing between two values is just called a "periodic point of period two". This is one of the most famous routes to chaos: the so-called "period-doubling route".  Lots of fun to be had from the parabola:  let  `double f(double x) { return a * x * (1.0-x); } and set `a` to anything bigger than 4.

 Which illustrates the next point: fixed points are usually minima/maxima, and which minima/maxima you find depends on where you start.  Your algorithm does not do hill-climbing; instead it iterates, so who knows what it will find. Hill-climbing is a more stable algo for finding fixed points.  It can also be faster than iterating.

ngeiswei:
Yeah, a fixpoint should be used only when it exists. Anyway, I've decided to move it the ure repo instead where it is used.

linas:
my apologies if I once again sounded bombastic. so it goes.

ngeiswei:
Maybe, but I agree that it's better to move it to cogutil if it turns out to be really often used, otherwise, given its simplicity its not worth the trouble.

And just to be clear although the unit test was using float, my use case is on discrete data structures where I know a fix-point exists, the choice of unit test was just not very clever.

PR #217, opened 2019-12-31
Bug-fix missing parallel support.

linas:
The CMakefile calls this `HAVE_PARALLEL_ALGORITHM` not `HAVE_PARALLEL_STL`

### DIFF
--- a/opencog/util/oc_omp.h
+++ b/opencog/util/oc_omp.h
@@ -46,7 +46,7 @@
 //! compile with multithread support (!)
 #if defined(CYGWIN)
 #elif defined(__APPLE__)
-#elif HAVE_PARALLEL_STL
+#elif HAVE_PARALLEL_ALGORITHM
 #define OC_OMP
 #endif
 


linas:
@ngeiswei -- please take a look. 

With stdc++17, I think pretty much everything you want is in std, and so OC_OMP isn't really needed, any more, except for .. what? clang and Apple? I only use OC_OMP in one place ... but .. well, I noticed something was wrong ...

linas:
Actually, this doesn't fix it either, somethings still wrong. `OC_OMP` is still not defined, even though cmake prints: `C++ library standardizes parallelism` ... so this is more of a bug-report than a pull-req .... 

Oh .. I see, there is also this:
```
IF (PARALLEL_STL_FOUND)
   ADD_DEFINITIONS(-DHAVE_PARALLEL_STL)
ENDIF (PARALLEL_STL_FOUND)
```
but that is not working ... arghh...

Issue #177, opened 2019-06-18
Upgrade to C++17?


ngeiswei:
Is it OK if we move to the C++17 standard?

That way no need to rely on boost for multi-threading programming goodies like `shared_mutex`. I must warn you though that it would likely require to increase gcc minimum version to 6, as according to https://en.cppreference.com/w/cpp/compiler_support#cpp17 gcc 5 has poor support for C++17.

linas:
On my devel  system, I'm running debian stable, which uses gcc version 6.3.0   On my data-processing LXC containers, I'm using ubuntu 16.04 which has gcc-5.4.0  and  would rather not disrupt those for the next few weeks or a bit more. It took me the last month to get everything stable and they're finally running and producing results, so I'm nervous about downtime. But maybe in a week or two...

linas:
Also should poll the HK labs.

ngeiswei:
OK, meanwhile I'll use boost, since C++17 provides pretty much a drop-in replacement for it, shouldn't be too much of a head-heck later on.

linas:
Do you ant to use these mutexes for atomspace, or for other repos (moses/as-moses/opencog)?  I only need the atomspace in these old containers; I could simply avoid pulling a new cogutils. I want to retain the ability to pull the newest atomspace, because I keep tripping over bugs in it.  I have no other dependencies...

ngeiswei:
Those mutexes are for the URE. MOSES uses boost mutexes and they seem rather untroublesome.

linas:
OK. Well, anyway, double-check your thinking to make sure you really need these. Other code in the atomspace seems to get by with just a plain mutex, or a recursive mutex.  You might be setting yourself up for trouble by using  shared mutex. Think about how to simplify the design, so that you don't need one.

Also, recall, FYI, that cogutils has thread-safe queues(fifo) and stacks(lifo) that are nearly identical to the std:: versions thereof.

ngeiswei:
Yes, I might use plain mutex for now, it's just that I've learned while working on multithreaded MOSES that shared mutex on read can improve performance, though I forgot the numbers and foolishly didn't document that is a diary.

linas:
[...]

PR #219, opened 2020-01-08
Move to C++17 (and set minimum gcc version to 6.0)

ngeiswei:
There does not seem to any porting required, tested repos (cogutil, atomspace and ure) compile and pass unit tests on my machine.


ngeiswei:
Fix issue #177

PR #222, opened 2020-02-27
Improve option to string by ignoring meaningless decimals

ngeiswei:


ngeiswei:
AS-MOSES fails, due to not being adjusted to the new atom API.

Will fix that, but meanwhile it's clear nothing fails because of that commit so I'm merging.

PR #223, opened 2020-04-01
Pass const ref in pool::give_back

ngeiswei:
The resource is copied, it has thus no reason not to be passed by
const.


ngeiswei:
Needless to say this limited the calling options, thus that PR.

PR #225, opened 2020-04-15
Ubuntu 20.04 binutils API fix

ntoxeg:
This is one of the three opened pull requests opened among OpenCog projects that changes and fixes a few scripts such that you can build and use OpenCog on the new LTS Ubuntu 20.04, coming out this April.

Unfortunately, there is a small fix that I had to apply to Cogutil, so these changes **are not** backwards-compatible.

What changed here is that a few procedures (namely `bfd_get_section_vma`, `bfd_get_section_flags` and `bfd_section_size`) changed their names and/or signature.

Sorry for the reformatting, I couldn't get clang-format to output the same style.

Again, this is not backwards-compatible so I'm just leaving this here in case some day you want to move to 20.04 so perhaps this will save you time.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,8 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
+Language: Cpp
+BreakBeforeBraces: Linux
+IndentCaseLabels: true
+IndentPPDirectives: BeforeHash
+ColumnLimit: 100
+AllowShortIfStatementsOnASingleLine: true
--- a/opencog/util/backtrace-symbols.c
+++ b/opencog/util/backtrace-symbols.c
@@ -49,64 +49,59 @@
 
 #if defined(HAVE_BFD) && defined(HAVE_IBERTY)
 
-#include "backtrace-symbols.h"
+    #include "backtrace-symbols.h"
 
-#define fatal(a, b) exit(1)
-#define bfd_fatal(a) exit(1)
-#define bfd_nonfatal(a) exit(1)
-#define list_matching_formats(a) exit(1)
+    #define fatal(a, b) exit(1)
+    #define bfd_fatal(a) exit(1)
+    #define bfd_nonfatal(a) exit(1)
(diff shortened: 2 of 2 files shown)


ntoxeg:
I forgot to mention that the tests pass on my new container (besides Python complaining that Cxxtest uses "is" instead of "==").

.clang-format
```diff
@@ -0,0 +1,8 @@
+BasedOnStyle: LLVM
+IndentWidth: 4
```
linas, line 2: FWIW, most comments are formated so the line length stays under 80 columns ... when indent is 3

linas: those are the ones with tab indents ... when space indents are used, many files use 4 spaces...

ntoxeg:
- `bfd_get_section_vma` renamed to `bfd_section_vma`
- `bfd_get_section_flags` renamed to `bfd_section_flags`
-  `bfd_section_size(abfd, section)` to `bfd_section_size(section)`

I guess this could be conditionally compiled but I don't know how to do that without adding some preprocessing with CMake (insert version into the file where it's checked against).

.clang-format
ntoxeg: Nowadays people argue that it is better to wrap at something larger than 80 because people have larger, and particularly wider screens, so it makes sense to relax this constraint somewhat. I do not want to argue for this though, I can change it to your liking.

linas:
merging, pull req #226 provides the backwards-compat fix

.clang-format
linas: I don't really care, my editor ignored clang-format ... my screens are larger, but that doesn't mean I have more columns ... I just have LOTS more windows ... all 80 columns wide.  That way, I don't have to flip around, everything is there in front of me.

linas:
Thanks, BTW.

manisabri:
I'm on 20.04 trying to build the the master now and the issue still exists

PR #227, opened 2020-05-04
Fix cancellation bugs in conncurrent queue/set

linas:
Apparently, no one ever closed a queue before. The code was buggy...


linas:
Err, no. The original design was fine. Late-night mistake...

PR #230, opened 2020-07-12
All new zipf unit tests

linas:
Tighten and expand the unit tests.


linas:
Merging. Atomspace build is failing due to haskell  breakage.

PR #233, opened 2020-07-15
Set RPath for MAC OSX

Habush:
This fixes linker issue on OS X because CMAKE_MACOSX_RPATH is not set by default.


linas:
@ngeiswei circle-ci is failing to even start .. any clue why?

cmake/OpenCogLibOptions.cmake
```diff
@@ -24,3 +24,9 @@ SET(CMAKE_SKIP_BUILD_RPATH FALSE)
 SET(CMAKE_BUILD_WITH_INSTALL_RPATH FALSE)
 SET(CMAKE_INSTALL_RPATH "${CMAKE_INSTALL_PREFIX}/lib/opencog")
 SET(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
+
+#This fixes linker issue on OS X
+#See https://cmake.org/cmake/help/v3.0/policy/CMP0042.html
+if (${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+	set(CMAKE_MACOSX_RPATH 1)
+endif()
```
linas, line 32: There's a missing new-line at the end-of-lfile (that's what that red circle means). Git/github will work with missing end-of-file markers, but will complain about them. Your editor is configured to always drop it (it might be a macos thing, not sure)

.. also, don't you want to put this at the top of the file, not the bottom? Before the other RPATH settings take effect?

ngeiswei:
I just pushed some changes, circleci worked. I don't know what could cause the problem in particaulr PR. @Habush you might want to rebase your branch onto the master, I don't expect it to fix it, but we never know.

Issue #236, opened 2020-08-16
Breakage with Binutils 2.35 (or maybe 2.34?)


mhatta:
Looks like `#ifdef bfd_section_flags` trick in opencog/util/backtrace-symbols.c doesn't work anymore with Binutils 2.35 (or did it work with 2.34?)

`#ifndef bfd_get_section_flags` works, but I'm not sure it's the way to go.  If it's okay, I can provide a PR.

linas:
Provide a pull req.

I recently saw some generic library that does stack traces, .. I've already forgotten it's name and can't find it instantly .. the long term solution might be to port to that.

Stack-tracing is a debugging tool, it mostly does not affect ordinary users, and I use it only as a prelim to taking a closer look with gdb ... anyway, crashes are rare...

EliasTsegaw:
https://wiki.gentoo.org/wiki/Binutils_2.34_porting_notes/undefined_reference_to_bfd_get_section_*

will this be a solution It Worked for me when installing on ubuntu 20.04 just removing bfdh

AmeBel:
@linas I tested https://github.com/opencog/cogutil/pull/237 with binutils 2.34 (on ubuntu 20.04) and 2.30 (on ubuntu 18.04) and this repo builds and tests passes. What failures are you referring to  @ https://github.com/opencog/cogutil/pull/237#issuecomment-675027698 ?

linas:
Hi @AmeBel  -- there seem to be two incompatible versions of binutils, one version has an extra argument in a handful of calls; the other version does not. The current code tries to handle both versions with some #ifdef magic.  The problem with #237 is that it defeats the #ifdef by making both version work the same way. That's why it's "obviously wrong".

PR #237, opened 2020-08-17
Fix breakage with Binutil 2.35 (#236)

mhatta:
Seems "#ifdef bfd_section_flags" doesn't work anymore with Binutils 2.35.  "#ifndef bfd_get_section_flags" should work.


linas:
OK, this is obviously wrong. I don't know what changed, but this pull req breaks all earlier binutils versions.

PR #238, opened 2020-08-17
Got a compile error for some function calls from bfd.h...

liamwhittle:
I've got no idea what any of this code does but these changes allowed me to compile from source and pass all tests.
¯\_(ツ)_/¯


linas:
Yeah, this might be the same as the #236 and #237 --the problem was that #237 breaks all older versions of binutils. Some kind of alternative way of detecting the newer bfd version is needed.

PR #240, opened 2020-11-12
Fix for Binutils API problems

### LINKED ISSUE #236: Breakage with Binutils 2.35 (or maybe 2.34?)

ntoxeg:
Fixes #236 for me, on Ubuntu 20.04. I didn't check for older versions but now the checks are more exhaustive. I took this idea from another project I found where they had a similar issue and ported it to CMake.

### DIFF
--- a/opencog/util/CMakeLists.txt
+++ b/opencog/util/CMakeLists.txt
@@ -111,6 +111,27 @@ IF (HAVE_BFD AND HAVE_IBERTY)
 		)
 	endif()
 
+	# Check if bfd_section_vma takes one or two arguments.
+	check_c_source_compiles(
+		"#include <bfd.h>
+ 		int main(int argc, char** argv) {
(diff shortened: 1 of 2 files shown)


ntoxeg:
CogServer and AS-MOSES failed - it seems unrelated.

linas:
Thanks!  Seems like the code is correct, and it builds on my machine. Unfortunate that the solution had to be so excessively complicated ;-(

linas:
Oh, sorry also for long delay. I was out playing hooky.

PR #243, opened 2021-01-19
Fix boost compile error

ngeiswei:
Explicitly use `boost::placeholders` in order to compile under boost 1.75. Alternatively I could have silenced the error, but this will likely be mandatory at some point so better do it right now. It requires to upgrade boost requirement to boost 1.60, which is 5 year old, so I think it is more than acceptable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,10 +100,10 @@ ENDIF (CYGWIN)
 SET(Boost_USE_STATIC_LIBS OFF)
 SET(Boost_USE_MULTITHREADED ON)
 # The minimum boost version required for building.
-SET(MIN_BOOST 1.46)
+SET(MIN_BOOST 1.60)
 
 # Required boost packages
-# 1.46 is minimum for required filesystem support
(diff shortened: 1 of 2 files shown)


ngeiswei:
The cogserver fails

https://app.circleci.com/pipelines/github/opencog/cogutil/49/workflows/9085a0c4-640d-4808-914a-bfcf3261592d/jobs/1670

which as become quite usually these days. If someone knows what is wrong it would be great if this someone could step in.

ngeiswei:
I'm merging, it's clear the problem with cogserver isn't caused by that PR.

linas:
> cogserver fails

actually, its Cython that fails, with an unresolved symbol. This unit test passes on my machine. This suggests the cython is loading shared libraries out of order and/or is failing to make a symbol resolution pass after loading.  Is this reproducible, or intermittent?  If it is intermittent, perhaps different runs are loading libraries at different addresses. Perhaps cython is accessing libraries before loading and initialization has completed ... ?

Issue #239, opened 2020-09-16
GCC version 6 is not enough to build Atomspace, at least 7 or later is required


vsbogd:
Cogutil macro checks that GCC is 6 or later:
https://github.com/opencog/cogutil/blob/d1f3df098824e14ed8e1a74933e7aeab2ea022fb/cmake/OpenCogGccOptions.cmake#L5-L9
but `std::set::merge` method which is used in `atomspace` code: 
https://github.com/opencog/atomspace/blob/a633fe2b66cbd9c83cb2d71517a19cc924758703/opencog/atoms/container/JoinLink.cc#L139
was introduced in GCC starting from version 7.

ngeiswei:
I have personally no problem incrementing the required version of gcc. Note that Debian 9 still uses gcc 6.

linas:
This implies that Debian 9 won't compile opencog ...

linas:
Pull req #245 bumps version to 7.0

linas:
Closing; fixed in #245  I set it to 7.0 but I think its safe to go to 7.5 even, as ubuntu 18.04 LTS has 7.5 in it.  Open a new bug if you want something else.

PR #245, opened 2021-02-27
Bug #239 - minumum compiler version is 7.0

linas:
Fix for bug #239


linas:
Hmmm ... asmoses failed. But I doubt it is due to this pull req. Is it normal for asmoses to fail? @ngeiswei any idea?

AmeBel:
asmoses has added a dependency 
https://github.com/opencog/asmoses/commit/3d84c524de32a7e241347abd4e7a2e990256ca11

linas:
thanks @AmeBel  fixing this in #248

Issue #244, opened 2021-02-27
Unknown CMake command "add_compile_definitions".


linas:
While refreshing on an Ubuntu bionic (18.04 LTS) system, I got the error
```
CMake Error at opencog/util/CMakeLists.txt:72 (add_compile_definitions):
  Unknown CMake command "add_compile_definitions".
```

Apparently, this requires a newer version of CMake? Which version? Currently, the very first line of `CMakeLists.txt` is 
```
CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12.2)
```
 but apparently, this is not enough?  Looking at the blame log, this shows that pull req #240 added this code, and it was provided by @ntoxeg and merged by me (apparently, without checking if it worked on bionic ...)

Any clue for what it would take to get this to work on bionic?

linas:
Apparently, this first appears in cmake version 3.12 while ubuntu 18.04 has only cmake version 3.10

ntoxeg:
The older way of doing this is to use `add_compile_options` and explicitly state the option. Another way is to have newer CMake installed - besides obvious (and undesirable I’m assuming) installation from source it should also be installable with Snap like this: `sudo snap install cmake —classic`.

AmeBel:
See https://apt.kitware.com/ for installing the latest cmake version on ubuntu.

linas:
fixed in #246

PR #246, opened 2021-02-27
Bug #244 -- allow older cmake versions

linas:
Ubuntu 18.04 does not have a new-enough CMake


linas:
This is failing cause asmoses won't build ... I doubt this is because of cogutils ... what's up with asmoses?Any comments, @ngeiswei @Eman22S @kasimebrahim

PR #248, opened 2021-03-04
Try to fix build break via asmoses dependency.

linas:
According to
https://github.com/opencog/asmoses/commit/3d84c524de32a7e241347abd4e7a2e990256ca11

PR #249, opened 2021-03-05
Add ure dependency to asmoses in circleci

ngeiswei:


ngeiswei:
Complementary fix for #248

ngeiswei:
asmoses still fails for a different reason, it's clear that fix however had the intended effect so I'm merging for now.

PR #250, opened 2021-03-05
We need to install asmoses before running the unit tests

ngeiswei:
Ideally we should not need that but for some reason escaping my
understanding we do.


ngeiswei:
It fails due to the cogserver, which it has for a while, but not from asmoses anymore, so I'm merging.

PR #252, opened 2021-08-08
Set thread name on logger thread

linas:
This should simplify debugging.


linas:
Merging. One asmoses unit test is failing, and it's not because of this.

PR #255, opened 2021-09-08
Remove unused functions.

linas:
I beleive these are unused anywhere in opencog. Just remove them.
I think that half the other portability wrappers are also unused;
that needs to be a later commit.


linas:
cleanup suggested by opencog/atomspace#2842

PR #257, opened 2021-09-13
Add platform specific code for setting thread names.

Habush:
This PR adds the code to set the name of the current running thread in a platform-dependent way. This can be used for fixing issue opencog/atomspace#2832.

Also adds the `-fopenmp` flag for clang compilers.


linas:
Heh. The build failed.

What platform is this for?

Habush:
It is for Linux and Apple.

PR #259, opened 2021-11-10
Fix guile finder

xywei:
I have a `libguile.h` under `/usr/include`, which belongs to 1.8.8 version of guile. In this case, `NO_DEFAULT_PATH` needs to be set for it to find the correct header file under `/usr/include/guile/3.0`.

### DIFF
--- a/cmake/FindGuile.cmake
+++ b/cmake/FindGuile.cmake
@@ -32,6 +32,7 @@ FIND_PATH(GUILE_INCLUDE_DIR libguile.h
 
 	/usr/include
 	/usr/local/include
+	NO_DEFAULT_PATH
 )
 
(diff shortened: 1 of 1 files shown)


linas:
Huh. I merged this, because I think it's fine .. but I'm confused. At line 33, two lines above your change, there is an explicit `/usr/include` in the path. Doesn't this cause the 1.8.8 `libguile.h` to be found?

xywei:
This is because `/usr/include` is among the default paths, which [take precedence over all the specified paths](https://cmake.org/cmake/help/latest/command/find_path.html).

linas:
> which take precedence

Thanks!  I guess I need more coffee now.

PR #267, opened 2023-09-14
Use after free

linas:
Sub out code that is causing a compiler warning of use-after-free.
Only moses/asmoses uses this code. It appears to be bit-rotted.


linas:
It appears that this patch breaks 27 out of 45 unit tests in as-moses, at least on ubuntu 2020 LTS

linas:
solution is to remove tree.c as in pull req #281

PR #274, opened 2024-06-24
Parametrising location of ld.so.conf files.

gl-yziquel:
Currently, the installation of ld.so.conf file related to cogutil occurs in /etc. This is not only not portable, as mentionned in the cmake file where this is defined, but moreover requires root privileges. The current PR allows to use -DCMAKE_CONF_PREFIX=/home/mini-me/.config to ensure that this file is deployed locally in somewhere else than /etc. This behaviour is not standard, and these files are not picked up by default. A bit of shell scripting can however make that happen and allow to installation of cogutils to work in userspace without root privileges.

Some script like that in a shell initialisation should do the trick:

```bash
if [[ -d ~/.config/ld.so.conf.d ]]; then
    for conf in ~/.config/ld.so.conf.d/*.conf; do
        export LD_LIBRARY_PATH=$(cat "$conf" | tr '\n' ':')$LD_LIBRARY_PATH
    done
fi
```

It is unclear, however, to me, whether such a snippet is currently necessary for installation of opencog on top of cogutil.

### DIFF
--- a/etc/CMakeLists.txt
+++ b/etc/CMakeLists.txt
@@ -5,4 +5,5 @@ file(WRITE ${CMAKE_CURRENT_BINARY_DIR}/opencog.conf
 	"${CMAKE_INSTALL_PREFIX}/lib${LIB_DIR_SUFFIX}/opencog\n")
 
 # Well, this isn't portable, but it works for Linux
-install(FILES ${CMAKE_CURRENT_BINARY_DIR}/opencog.conf DESTINATION "/etc/ld.so.conf.d")
+set(CMAKE_CONF_PREFIX "/etc" CACHE PATH "Directory for configuration files")
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/opencog.conf DESTINATION "${CMAKE_CONF_PREFIX}/ld.so.conf.d")


linas:
Thanks. I merged the patch, it seemed fine.  Please let me know what your `build/etc/opencog.conf` contains. I assume it becomes `/home/mini-me/somewhere/lib/opencog` or something similar?

gl-yziquel:
> Thanks. I merged the patch, it seemed fine. Please let me know what your `build/etc/opencog.conf` contains. I assume it becomes `/home/mini-me/somewhere/lib/opencog` or something similar?

It contains:

```console
/home/mini-me/.local/lib/opencog
```

Which works fine with what the rest of opencog I've built on top. For now...

linas:
> XDG freedesktop conventions,

Ah, OK. If you end up with more patches, send them in.

I've solved the "multiple versions" problem by using LXC and docker. I'm not sure which I like better.

gl-yziquel:
> I've solved the "multiple versions" problem by using LXC and docker. I'm not sure which I like better.

Yeah. I use lxc with incus. But I don't like using it for such tooling but only for infrastructure purposes.

I really do not like docker.

linas:
Heh. I thought I hated docker, too, but if I have to show a demo for someone else, its fairly foolproof.

gl-yziquel:
I grant you that.

PR #275, opened 2025-05-28
Boost minor cleanup

linas:
Remove some pointless invocations of boost


linas:
Never mind. as-moses broke due to the std::bind conversion :-(

PR #277, opened 2025-06-16
Fix macOS build errors

issacnitin:
Currently, installing atomspace fails in MacOS if guile is installed using homebrew (`brew install guile`)

Also fixed compilation error

`
/usr/local/include/opencog/util/random.h:137:20: error: implicit instantiation of undefined template 'std::basic_stringstream<char>'
  137 |         std::stringstream ss;
`


linas:
Thanks!

